PR #5, opened 2021-08-13
fix: better error handling for JS/WASM APIs

kitsonk:
This fixes an issue with `createGraph()` that would panic if a module specifier was bad.  It also adds several tests for checking for error conditions in the JS/WASM APIss.


test.ts
```diff
@@ -46,11 +48,59 @@ https://example.com/a.ts (32B)
...
+});
+
+Deno.test({
+  name: "createGraph() - load throws",
+  async fn() {
+    const graph = await createGraph("file:///a/test.ts", {
+      load() {
+        throw new Error("something bad happened");
+      },
+    });
```
bartlomieju, line 94: It's a bit surprising that error thrown in `load` function is not re-thrown from `createGraph`, but I guess that's by design and not mistake?

bartlomieju (approved):
LGTM, but as noted below the API is a bit surprising to me

dsherret (approved):

test.ts
kitsonk: Any errors thrown by load get stored in the module graph slot.  So if you were to do a `graph.get("file:///a/test.ts")` the error would be "re-thrown" at that point.

bartlomieju: Ah, thanks, I forgot about that

PR #6, opened 2021-08-13
refactor: remove wee_alloc

kitsonk:
Looks like wee_alloc might be causing problems on deploy, it was already of doubtful value, so want to get rid of it.


bartlomieju (approved):
LGTM, it was actually Luca's suggestion to remove it ;)

PR #8, opened 2021-08-17
fix: content type parsing

kitsonk:
Properly handles deriving the media type from content-types that include the content encoding.  Also updates swc.


bartlomieju (approved):
LGTM, good fix

PR #20, opened 2021-08-31
chore: update wasm loader

kitsonk:
Aligns with the changes to `deno_doc`.


bartlomieju (approved):
LGTM

_build.ts
```diff
@@ -101,22 +101,40 @@ console.log(
...
+const loader = `let wasmInstantiatePromise;
+switch (wasm_url.protocol) {
+  case "file:": {
+    const wasmCode = await Deno.readFile(wasm_url);
+    wasmInstantiatePromise = WebAssembly.instantiate(wasmCode, imports);
+    break;
+  }
+  case "https:":
+  case "http:": {
+    if ("permissions" in Deno) Deno.permissions.request({ name: "net", host: wasm_url.host });
```
dsherret, line 113: Should we do the same thing with the `readFile` above? I noticed it was removed below for some reason? (maybe accidentally?)

dsherret (approved):
LGTM. Just that one read permissions question.

lib/deno_graph.js
```diff
@@ -208,7 +208,7 @@ function makeMutClosure(arg0, arg1, dtor, f) {
 }
```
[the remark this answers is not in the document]
bartlomieju: I agree with that, it always takes me a moment to figure out that function names in this file can't be written be a human.

_build.ts
kitsonk: It wasn't accidental.  I realised that the only thing I could do was ask for read permissions concisely and asking for permission to everything felt "wrong", but now that you bring it up again, I realise I can pass the `wasm_url` as the `path` and only request read permission for that.

PR #21, opened 2021-09-02
fix: export Resolved

kitsonk:


src/lib.rs
```diff
@@ -1,189 +1,184 @@
...
+          Some(parser) => parser,
+          None => &mut default_parser,
+        },
+      );
+      builder.build().await
+    }
+
+    /// Parse an individual module, returning the module as a result, otherwise
+    /// erroring with a module graph error.
+    #[cfg(feature = "rust")]
```
bartlomieju, line 67: Seems superfluous?

src/lib.rs
```diff
@@ -1,189 +1,184 @@
...
+        },
+      ) {
+        ModuleSlot::Module(module) => Ok(module),
+        ModuleSlot::Err(err) => Err(err),
+        _ => unreachable!("unreachable ModuleSlot variant"),
+      }
+    }
+
+    /// Parse an individual module from an AST, returning the module.
+    #[cfg(feature = "rust")]
```
bartlomieju, line 94: Ditto

bartlomieju (approved):

src/lib.rs
kitsonk: Yeah, it is... missed it.

PR #36, opened 2021-09-29
feat: changes to integrate into Deno CLI

kitsonk:
This PR contains several features/changes to make integrating into Deno CLI possible:

- Support multiple root specifiers when building a module graph. There are a couple use cases in the CLI where we have multiple roots (like when globbing tests), so it is unavoidable change `root` to `roots`.  JavaScript/TypeScript handles "overloading" of function signatures easily, so this is a non-breaking change for `createGraph()` but does break the Rust `create_graph()`.  Also the serialization is `roots` instead of `root` and will always be an array of module specifiers.
- Add the concept of preferring types when resolving dependencies out of the graph.  This is important when type checking a graph versus trying to "run" a graph, which CLI needs to do both.
- Several methods on the graph that allow access to the data in the graph in a supportable way, like `.modules()`, `.errors()`, `.specifiers()`, and `.valid()` without tightly coupling to the structure of the graph.

This should be reviewing in context of the changes to CLI (and import_map and deno_doc).


src/graph.rs
```diff
@@ -332,12 +529,90 @@ impl ModuleGraph {
...
+      match get_module(specifier) {
+        Ok(Some(module)) => {
+          if let Some((_, Resolved::Specifier(specifier, _))) =
+            &module.maybe_types_dependency
+          {
+            validate(specifier, seen, get_module)?;
+          }
+          for dep in module.dependencies.values() {
+            if !dep.is_dynamic {
+              match &dep.maybe_code {
```
dsherret, line 565: It seems the code here and for `maybe_type` are the same? Would it make sense to eliminate this duplication?

src/graph.rs
```diff
@@ -332,12 +529,90 @@ impl ModuleGraph {
...
+                Resolved::Specifier(specifier, _) => {
+                  validate(specifier, seen, get_module)?
+                }
+                Resolved::Err(err, span) => {
```
dsherret, line 569: It might be nice to store the `span` in `ResolutionError` or do something so that we can change `maybe_code` to a `Result` and we could just do `dep.maybe_code?` above to return the error. That would also help reduce duplication here mentioned in my comment above because we'd only after to do one if let statement that calls `validate` in the body for both `maybe_code` and `maybe_type`.

src/graph.rs
```diff
@@ -232,37 +265,92 @@ impl ModuleSlot {
...
+    self
+      .module_slots
+      .get(&specifier)
+      .map_or(false, |ms| matches!(ms, ModuleSlot::Module(_)))
+  }
+
+  /// Returns any errors that are in the module graph, along with the associated
+  /// specifier.
+  #[cfg(feature = "rust")]
+  pub fn errors(&self) -> Vec<(&ModuleSpecifier, &ModuleGraphError)> {
```
dsherret, line 335: Would it be better if we just stored `ModuleSpecifier` within `ModuleGraphError`? Then consumers could do something like `error.specifier()` and we could eliminate this tuple here and the usages elsewhere. I think it would make consuming code easier to understand

Side comment: never liked `error.0` and `error.1`... C# at least has named tuples, but these Rust tuples aren't ideal for a public API.

src/graph.rs
```diff
@@ -288,32 +376,87 @@ impl ModuleGraph {
...
+  /// The `prefer_types` flags indicates if a type dependency is preferred over
+  /// a code dependency. If `true`, a type dependency will be returned in favor
+  /// of a code dependency. If `false` a code dependency will be returned in
+  /// favor of a type dependency. The value should be set to `true` when
+  /// resolving specifiers for type checking, or otherwise `false`.
   pub fn resolve_dependency(
     &self,
     specifier: &str,
     referrer: &ModuleSpecifier,
+    prefer_types: bool,
```
dsherret, line 429: Kind of makes me sad when there's a boolean parameter and I can't tell what it is from the call site in a code review. Just makes it easier not to catch a mistake on a code review, but this is a failing of code review tools.

I wonder if it would make sense if we change this to (I'm just throwing some ideas out there, not really suggestions):

```rust
.resolve_dependency_prefer_types(specifier, referrer)
// then a corresponding
.resolve_dependency_prefer_code(specifier, referrer)
```

Or maybe even a struct that's returned that returns both... then people could call:

```rust
.resolve_dependency(specifier, referrer).prefer_types()
```

dsherret (approved):
LGTM! I only have nitpicks and questions.

src/graph.rs
```diff
@@ -232,37 +266,86 @@ impl ModuleSlot {
...
 #[derive(Debug, Serialize)]
 pub struct ModuleGraph {
-  pub root: ModuleSpecifier,
+  pub roots: Vec<ModuleSpecifier>,
   #[serde(skip_serializing)]
-  maybe_locker: Option<Rc<RefCell<dyn Locker>>>,
-  #[serde(serialize_with = "serialize_modules")]
-  pub(crate) modules: BTreeMap<ModuleSpecifier, ModuleSlot>,
-  redirects: BTreeMap<ModuleSpecifier, ModuleSpecifier>,
+  maybe_locker: Option<Rc<RefCell<Box<dyn Locker>>>>,
```
bartlomieju, line 297: `Rc<RefCell<Box<dyn Locker>>` looks like antipattern and I'm surprised clippy doesn't complain. I think you should be able to store it in `Rc<RefCell<dyn Locker>>` instead.

src/graph.rs
bartlomieju: I agree with David

src/graph.rs
```diff
@@ -288,40 +372,179 @@ impl ModuleGraph {
...
     while let Some(specifier) = self.redirects.get(redirected_specifier) {
+      if !seen.insert(specifier) {
+        eprintln!("An infinite loop of redirections detected.\n  Original specifier: {}", specifier);
+        break;
+      }
       redirected_specifier = specifier;
+      if seen.len() > 5 {
+        eprintln!("An excessive number of redirections detected.\n  Original specifier: {}", specifier);
+        break;
+      }
```
bartlomieju, line 426: FYI in CLI we have redirect limit set to 10

kitsonk: Not where I borrowed the code from: https://github.com/denoland/deno/blob/11acdf1ea8b2978dcacad37b386f45e43a1f2748/cli/module_graph.rs#L1694-L1697

src/graph.rs
kitsonk: Yes, I was hacking away on this so much that I didn't realise what I ended up with.

src/graph.rs
bartlomieju: I see, I was referring to this one: https://github.com/denoland/deno/blob/11acdf1ea8b2978dcacad37b386f45e43a1f2748/cli/file_fetcher.rs#L513

src/graph.rs
kitsonk: I will think about it, but it is explicit and clear in an editor.

I am not convinced about this suggestion though, having long duplicated methods names or a complicated return structure doesn't make it simple to me.  The compiler would make anyone be explicit about what they want when consuming the API, even if their editor didn't make it clear, which it does.

src/graph.rs
kitsonk: Yes, agreed.

src/graph.rs
kitsonk: Yeah, agreed.

src/graph.rs
kitsonk: Done.

src/graph.rs
```diff
@@ -288,32 +376,87 @@ impl ModuleGraph {
...
-      if let Resolved::Specifier(specifier, _) = &dependency.maybe_type {
-        Some(specifier)
-      } else if let Resolved::Specifier(specifier, _) = &dependency.maybe_code {
-        Some(specifier)
+      let (maybe_first, maybe_second) = if prefer_types {
+        (&dependency.maybe_type, &dependency.maybe_code)
+      } else {
+        (&dependency.maybe_code, &dependency.maybe_type)
+      };
+      if let Resolved::Specifier(specifier, _) = maybe_first {
```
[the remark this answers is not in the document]
kitsonk: Done.

src/graph.rs
kitsonk: It is complicated to eliminate the duplicate code here, because of nested function, so it would be another nested function, and I think it would sort of look like a similar number of LOCs.  I've added a TODO for now.

src/graph.rs
```diff
@@ -194,19 +223,23 @@ pub struct Module {
   )]
   pub maybe_types_dependency: Option<(String, Resolved)>,
   pub media_type: MediaType,
+  #[serde(skip_serializing)]
+  pub parsed_source: ParsedSource,
```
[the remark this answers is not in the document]
kitsonk: I suspect we can eliminate it in a follow up PR.  Added a TODO.

src/graph.rs
kitsonk: Hmmm... no, this doesn't work at the other end... and I can't cast it to a locker unless the trait is sized, but trying to add sized to a trait is beyond me and causes a whole lot of other issues.

PR #61, opened 2021-10-30
fix: Load static deps of dynamic deps as dynamic

nayeemrmn:
```js
// file:///foo.js (root)
await import ("file:///bar.js");

// file:///bar.js
import "file:///baz.js";

// file:///baz.js
console.log("Hello, world!");
```
Makes `file:///baz.js` load with `self.load(specifier, true)` instead of `self.load(specifier, false)`.


kitsonk (commented):
LGTM

We need to rebuild the WASM as part of this PR though and edits to the branch aren't allowed.

nayeemrmn:
Updated. I have allow edits checked, though.

kitsonk:
> Updated. I have allow edits checked, though.

Ok, odd, might have been a workflow problem on my side, as I was working off a codespace and sometimes it is odd.

kitsonk (approved):
LGTM

PR #62, opened 2021-11-01
feat: support identifying jsxImportSource pragma dependencies

kitsonk:
Ref: denoland/deno#8440

This PR adds the following features:

- parses out the `@jsxImportSource` pragma from a module
- adds the emitted module specifier as a dependency to the module
- Allows the loader trait to supply the specifier module, defaulting to `jsx-runtime` (this is needed because when specifying "dev" mode, the implicit module name changes to `jsx-dev-runtime`)

This, plus the concept of "imports" already in the graph should allow us to support the React 17 JSX transforms in CLI (and in theory Deploy) in the least "hacky" way possible.

For context, something like this:

```ts
/* @jsxImportSource https://esm.sh/preact */

export function A() {
  return <div>Hello Deno</div>;
}
```

will get emitted to something like this:

```js
import { jsx as _jsx } from "https://esm.sh/preact/jsx-runtime";
/* @jsxImportSource https://esm.sh/preact */
export function A() {
    return _jsx("div", { children: "Hello Deno" }, void 0);
}
```

And the parsing of the module would identify that `https://esm.sh/preact/jsx-runtime` would be what would be a runtime dependency and adds it to the dependencies of the graph module.


src/source.rs
```diff
@@ -63,6 +63,11 @@ pub trait Loader {
   fn get_cache_info(&self, _specifier: &ModuleSpecifier) -> Option<CacheInfo> {
     None
   }
+  /// An optional method which returns the JSX import source module which will
+  /// be appended to any JSX import source pragmas identified.
+  fn jsx_import_source_module(&self) -> &str {
+    "jsx-runtime"
+  }
```
dsherret, line 70: I'm not sure about this being on the loader trait. Seems more like an option that can be provided when creating a graph. That way someone doesn't need to provide a different loader based on what `jsx_import_source_module` string they want to use.

dsherret (approved):
LGTM

src/source.rs
kitsonk: Well, I don't want to start a slippery slope of config for the graph. Maybe it isn't the loader trait, but actually the resolver trait though now that I think of it and the lack of a resolver will default to this default.

kitsonk: Ok, I moved it to the Resolver trait and I like it better, as it removed it from `parse_module`.  It only really comes into play in very narrow situations, basically when you want to support the development mode for JSX transforms, which really is only a runtime thing, not really any other point in time. I really don't want to do it all.

dsherret: Yup, resolver trait is better!

dsherret (approved):
LGTM

PR #63, opened 2021-11-01
feat: Add ModuleGraph::get_referrer()

nayeemrmn:
Like [`GraphData::resolved_map`](https://github.com/denoland/deno/blob/v1.15.3/cli/proc_state.rs#L66), but that is unusually populated during V8 module load. This makes it available earlier.


kitsonk:
What problems does this solve?

nayeemrmn:
I need it to add referrer info here: https://github.com/denoland/deno/pull/12604/files#diff-0d33a1d418f18ec5f991ddde53721750aa7f2950e811f0d93836fd0a60ed5318R364. As we do at https://github.com/denoland/deno/blob/v1.15.3/cli/proc_state.rs#L566, but it should be done in `prepare_module_load()`.

kitsonk:
Not everything needs to go in the graph, I don't see how this information is usable outside of the very specific module loading in the CLI, which means it should reside there.  It was originally generated during prepare_module_load but @bartlomieju moved it.

The information makes little sense in the JSON export, like what would people use it for?  The other parts of the JSON structure make sense and are generally applicable of what a "module graph" would be.

I can't see why this can't be built from an already built graph, similar to the dependants logic in the lsp: https://github.com/denoland/deno/blob/main/cli/lsp/documents.rs#L427

nayeemrmn:
Referrer info is useful for any integration of `ModuleGraphError` IMO. What if instead of having this API we tracked them internally on the builder and appended `\n    at <referrer>` to the error messages in `ModuleGraphError::fmt()`?

nayeemrmn:
I've excluded `referrer_map` from the JSON.

nayeemrmn:
Changed API from `ModuleGraph::referrer_map()` to `ModuleGraph::get_referrer()`.

PR #67, opened 2021-11-08
fix: resolve_dependency considers synthetic imports

kitsonk:
When resolving the `jsxImportSource` for `tsc`, we need the ability to resolve dependencies from a synthetic module, so we can "find" the `jsxImportSource` configured from a set of compiler options.


lucacasonato (approved):

dsherret (approved):
LGTM

PR #69, opened 2021-11-10
feat: support analysis of `require()` statements

### LINKED ISSUE #58: Support CommonJs dependency analysis
swc already supports parsing this out (we just simply discard those nodes at the moment), but I think the best way we can identify them in the graph is to add a `is_require` to the resolved dependency and pass it in the `load()` method, like we do with `is_dynamic`.

I will work on a PR for it.  Makes me feel a bit 🤢 though 😉.

kitsonk:
Closes #58

The PR allows the `deno_graph` to statically analyse `require()` statements in modules and have them exist as part of the module graph.  There are some limitations, which are swc limitations:

- It statically looks for the `require()` symbol in code and does not track re-assignment.  Meaning it won't find dependencies like `const r = require; const b = r("b");`.
- It only detects static strings passed to the `require()` function, therefore things like `const m = "m"; require(m);` do not work.

This is a BREAKING CHANGE on the Rust side, but one that shouldn't impact any of the CLI functionality.  Essentially `Resolver::resolve()` and `Loader::load()` now take an extra argument which indicates what kind of dependency it is.  This allows consumers to be able to branch logic in code based on the "reason" the resolution or loading is occurring.  On the JavaScript side, it is a non-breaking change, since excess arguments are ignorable.


dsherret (commented):
Just did a quick skim through and it's looking good to me.

src/graph.rs
```diff
@@ -289,8 +290,55 @@ where
...
+  Require,
+  /// The dependency was identified via a pragma within the file
+  Pragma,
+}
+
+impl Default for DependencyKind {
+  fn default() -> Self {
+    Self::Import
+  }
+}
```
bartlomieju, line 314: Is it required to implement `Default` trait for this enum? It feels to me like a potential footgun that might lead to discrepancies by too liberal use of `Default::default()` somewhere in the call chain.

bartlomieju (approved):
LGTM, only one nitpick. I think we should wait on upgrade to `deno_ast` before landing this PR (https://github.com/denoland/deno_ast/issues/46)

bartlomieju:
@kitsonk is there anything I can help you to land this PR?

kitsonk:
It needs quite a bit of rebasing... let me take care of it.  We just didn't need it before, now there is a compelling need.

kitsonk:
Stale and uneeded.

PR #75, opened 2021-12-01
fix: Remove referrer from SpecifierError::fmt()

nayeemrmn:
This creates a redundant referrer in https://github.com/denoland/deno/blob/v1.16.3/cli/tests/testdata/095_cache_with_bare_import.ts.out. The `    at <referrer>` with line and column numbers is better.


kitsonk (approved):
LGTM

PR #78, opened 2021-12-13
feat: support import assertions

kitsonk:
This add "proper" import assertion support to `deno_graph`. It is going to be a breaking change (but a pretty light one).  This models it properly and should be able to accomodate CSS Modules in the future.

I haven't tried to integrate it into CLI yet, but it should be fairly straight forward.


lucacasonato (commented):
I don't think this is quite correct. The loader should not care about import assertions I think (as they are a runtime safety feature). The assertions should be validated by `deno_core` on the basis of the module type that the loader says the module is.

Import assertions also should never change the loading behavior as per spec, so a loader does not need to be aware of them.

Also I think this PR also may not be doing the right thing during concurrent loads of the same specifier with different assertions:

```js
import foo from "./file.json"; 
import foo from "./file.json" asserts { type: "json" };
```

kitsonk:
> Import assertions also should never change the loading behavior as per spec, so a loader does not need to be aware of them.

Fair point.  I will remove it from the loader trait.

> Also I think this PR also may not be doing the right thing during concurrent loads of the same specifier with different assertions

Grrr... You are right. I need to re-work that.

nayeemrmn:
> The assertions should only be validated by `deno_core` on the basis of the module type that the loader says the module is.

We want to detect import assertion failures in `deno cache`, right? It would be good to have them here as `ModuleGraphError`s (keeping the current loader API). Maybe it's `deno_core` that shouldn't process import assertions since it's closer to host-agnostic JS, whereas `deno_cli` is the host.

lucacasonato:
> We want to detect import assertion failures in `deno cache`, right? It would be good to have them here as `ModuleGraphError`s (keeping the current loader API). Maybe it's `deno_core` that shouldn't process import assertions since it's closer to host-agnostic JS, whereas `deno_cli` is the host.

No, this is a runtime validation error that should be surfaced at runtime IMO. We also don't error `deno cache` when you have a `new Worker` / `import()` with an invalid URL in your source.

lucacasonato:
And `deno_core` needs to be aware of module type anyway for JSON modules, because it needs to know how to construct a module record from the module source.

nayeemrmn:
> We also don't error `deno cache` when you have a `new Worker` / `import()` with an invalid URL in your source.

That's because those are dynamic. I don't see how assertions in general are runtime concepts. Assertions for static imports should be statically enforced and assertions for dynamic imports should be dynamically enforced.

kitsonk:
I agree with @lucacasonato, `deno cache` doesn't specifically care about if the module can be loaded or not.  They should be part of the dependency graph _if_ it looks like it is a valid import, but the real authority is `deno_core` which has to compare the media type of the module with the assertion type.

If you have something in the cache, and you try to import it is `import a from "a.json"` and it doesn't have the assertion it fails.

The import assertions for dynamic imports can in many cases only be determined at runtime, just like sometimes the import specifier can't be statically determined, so the graph cannot be authoritative about a module load error.

nayeemrmn:
It would be an awful shame if `deno bundle`/`deno compile` just tossed away type assertions.

> `deno cache` doesn't specifically care about if the module can be loaded or not.

> If you have something in the cache, and you try to import it is `import a from "a.json"` and it doesn't have the assertion it fails.

That's not a correct summation of what `deno cache` does. If you have an existing emit from `--no-check` but there are type errors, `deno cache` is supposed to catch that. Or if a file is already downloaded but there is an integrity error, `deno cache` catches that. I don't see why it wouldn't error on a statically identified module branch issue like a failed static type assertion. `deno cache` generally prepares things idempotently _and then_ validates what it has prepared.

> The import assertions for dynamic imports can in many cases only be determined at runtime, just like sometimes the import specifier can't be statically determined, so the graph cannot be authoritative about a module load error.

Currently we statically populate module slots for dynamic imports only when they have "statically analysable" specifiers. So to correctly support this, we should statically populate module slots for dynamic imports only when they have "statically analysable" specifiers + assertions. It's a completely logical extension to the magic we already do with dynamic specifiers.

bartlomieju:
Given that we only have 3 days left till 1.17 release I suggest we just alter `deno_graph` to not error out on `Json` module as part of the graph so I can actually wire up assertions and JSON module support in `deno_core`. We can then discuss if graph should validate assertions or not. I actually don't have strong opinion either way, the best we can do is validate it for static imports/exports anyway as assertions in dynamic imports can actually use variables instead of only using string literals.

nayeemrmn:
I agree, it would be best to do that minimal change for 1.17 and I'm happy to work on surfacing assertion errors in `prepare_module_load()` afterwards.

<details>
<summary></summary>

> the best we can do is validate it for static imports/exports anyway as assertions in dynamic imports can actually use variables instead of only using string literals.

Keep in mind, this statement applies to the dynamic import _specifier_ as well. We can take our well-established system for that and extend it to assertions 😉 i.e. don't statically visit a dynamic branch unless the specifier **as well as the assertions** are passed as simple literals.
</details>

bartlomieju:
> Keep in mind, this statement applies to the dynamic import specifier as well. We can take our well-established system for that and extend it to assertions 😉 i.e. don't statically visit a dynamic branch unless the specifier as well as the assertions are passed as simple literals.

Of course, that's true.

src/graph.rs
```diff
@@ -395,39 +466,71 @@ impl Module {
...
+    skip_serializing_if = "BTreeMap::is_empty",
+    serialize_with = "serialize_synthetic_dependencies"
+  )]
   pub dependencies: BTreeMap<String, Resolved>,
+  #[serde(flatten, skip_serializing_if = "Option::is_none")]
+  pub maybe_cache_info: Option<CacheInfo>,
+  #[serde(rename = "checksum", skip_serializing_if = "Option::is_none")]
+  pub maybe_checksum: Option<String>,
+  #[serde(
+    rename = "size",
```
bartlomieju, line 486: is this an intentional rename?

src/lib.rs
```diff
@@ -1228,10 +1419,78 @@ console.log(a);
...
+    let result = parse_module(
+      &specifier,
+      None,
+      Arc::new(
+        r#"
+    import a from "./a.json" assert { type: "json" };
+    await import("./b.json", { assert: { type: "json" } });
+    "#
+        .to_string(),
+      ),
```
bartlomieju, line 1443: Would be great to have a test for `export ... from ... assert {type: "json"};` and `await import(...", { assert: { type: variable } })`, but it's not crucial to land this PR

bartlomieju (approved):
LGTM, it seems all required info is provided in this PR

src/lib.rs
kitsonk: let me add those...

src/graph.rs
kitsonk: Yes. It is the same with ESModule.

src/graph.rs
```diff
@@ -949,11 +1049,30 @@ pub(crate) fn parse_module(
...
+  // here we check any known import assertions and add them to the graph,
+  // otherwise we just continue
+  match maybe_assert_type {
+    Some("json") if media_type == MediaType::Json => {
+      return ModuleSlot::Module(Module::Synthetic(Box::new(
+        SyntheticModule::new(
+          specifier.clone(),
+          MediaType::Json,
+          None,
+          Some(content),
```
nayeemrmn, line 1067: Why are JSON modules synthetic? Synthetic modules were previously side roots (e.g. config file) that brought in ambient deps like JSX import sources.

kitsonk: It has been refactored, there was no really distinction between the two and V8 creates them as "synthetic modules".  Now synthetic modules play the role of all modules that aren't ES modules, injected or otherwise, and the structure handles that they may or may not have source content.

nayeemrmn: Why not unify them with `Module` which could may or may not have a `parsed_source`? That seems a lot more sane. TS modules are also a special variant that need extra processing by the runtime, it doesn't make sense for those to be closer to ES modules than JSON modules are.

kitsonk: Integrating it into deno_doc and cli I am confident that we should keep JSON modules (and other asserted modules) as synthetic modules.  Injected sources modules are another class of synthetic modules.

There are clearly going to be refactors after this, like how we deal with JSON modules in bundling.  The fact they can "reside" in the graph, even if we refactor it later, is what we need to do for now.

PR #92, opened 2021-12-21
feat: support analysis of JSDoc imports

### LINKED ISSUE #91: graph does not analyze JSDoc imports
https://github.com/denoland/deno/blob/1eb78731eb57b1d0eb7c0ece97b2018c1724989e/cli/lsp/documents.rs#L1025-L1059

This is where it goes wrong. Because the dependencies list is empty, it fails to resolve the imported module.
I think think it might be possible to fix this without making changes to swc.
If you add a regular import at the top of a file, that fixes the jsdoc import:
```js
import {Bar} from "./Bar.js";

/** @type {import("./Bar.js").Bar} */
const foo = null;
```

_Originally posted by @jespertheend in https://github.com/denoland/vscode_deno/issues/588#issuecomment-998330173_

kitsonk:
Closes #91


src/ast.rs
```diff
@@ -370,4 +396,38 @@ mod tests {
...
+}
+"#
+      .to_string(),
+    );
+    let parser = DefaultSourceParser::new();
+    let result = parser.parse_module(&specifier, source, MediaType::TypeScript);
+    assert!(result.is_ok());
+    let parsed_source = result.unwrap();
+    let dependencies = analyze_jsdoc_imports(&parsed_source);
+    println!("{:?}", dependencies);
```
bartlomieju, line 431: Shouldn't there be an assert here instead of print? There's no expected list of deps in this test case

bartlomieju (approved):
LGTM, sans missing assertion

dsherret (approved):
LGTM other than the one test Bartek pointed out.

PR #93, opened 2021-12-21
feat: improve error messages related to import assertions

### LINKED ISSUE #82: Improve error message for wrong JSON imports
```
$ cat test.js
import * as data from "./data.json";
$ cargo run -- run -A test.js
error: An unsupported media type was attempted to be imported as a module.
  Specifier: file:///Users/biwanczuk/dev/deno/data.json
  MediaType: Json
    at file:///Users/biwanczuk/dev/deno/test.js:1:23
```

The error message should suggest to add `assert {type: "json}` to the import statement.

For reference Chrome gives following message:
```
Expected a JavaScript module script but the server responded with a MIME type of "application/json". Strict MIME type checking is enforced for module scripts per HTML spec.
```

CC @kitsonk

kitsonk:
Closes #82 

When importing a JSON module without an import assertion:

```
Expected a JavaScript or TypeScript module, but identified a Json module. Consider importing Json modules with an import assertion with the type of "json".
  Specifier: file:///a/a.json"
```

When importing something else that isn't supported:

```
Expected a JavaScript or TypeScript module, but identified a Wasm module. Importing these types of modules is currently not supported.
  Specifier: file:///a/e.wasm
```

When importing a non JSON module with a `"json"` import assertion:

```
Expected a Json module, but identified a JavaScript module.
  Specifier: file:///a/c.js
```

When using an unsupported import assertion type:

```
The import assertion type of "css" is unsupported.
```

When integrated into CLI, CLI will print out the associated location related to the error message.


src/graph.rs
```diff
@@ -175,12 +195,12 @@ impl fmt::Display for ModuleGraphError {
...
+      Self::InvalidSource(specifier, Some(filename)) => write!(f, "The source code is invalid, as it does not match the expected hash in the lock file.\n  Specifier: {}\n  Lock file: {}", specifier, filename),
+      Self::InvalidSource(specifier, None) => write!(f, "The source code is invalid, as it does not match the expected hash in the lock file.\n  Specifier: {}", specifier),
+      Self::InvalidTypeAssertion { specifier, actual_media_type, expected_media_type } => write!(f, "Expected a {} module, but identified a {} module.\n  Specifier: {}", expected_media_type, actual_media_type, specifier),
+      Self::UnsupportedMediaType(specifier, MediaType::Json) => write!(f, "Expected a JavaScript or TypeScript module, but identified a Json module. Consider importing Json modules with an import assertion with the type of \"json\".\n  Specifier: {}", specifier),
+      Self::UnsupportedMediaType(specifier, media_type) => write!(f, "Expected a JavaScript or TypeScript module, but identified a {} module. Importing these types of modules is currently not supported.\n  Specifier: {}", media_type, specifier),
```
bartlomieju, line 202: Tangential, but it's strange that these lines go over 80 columns

src/graph.rs
```diff
@@ -175,12 +195,12 @@ impl fmt::Display for ModuleGraphError {
...
+      Self::UnsupportedImportAssertionType(_, kind) => write!(f, "The import assertion type of \"{}\" is unsupported.", kind),
```
bartlomieju, line 203: FYI for this situation Chrome returns: `TypeError: "foobar" is not a valid module type.`, but this is fine for me too

bartlomieju (approved):
LGTM

src/graph.rs
kitsonk: We never wrap error messages, we allow whatever terminal to do wrapping.

bartlomieju: Yes, but the whole line of Rust code is over 80 lines, I'd expect it to wrap path after `=>` in braces and to make lines shorter.

kitsonk: oh, you mean `cargo fmt`? there are several scenarios now where `cargo fmt` doesn't wrap.  I am personally happy with it.

Issue #95, opened 2021-12-24
Add ability to have _type only_ and _code only_ graphs


kitsonk:
Currently, when building a graph, both type dependencies and code dependencies are analysed, because most consumers care about both of these types of dependencies, but when generating documentation (e.g. `deno_doc`) or just running code (e.g. Deploy/eszip) only parts of the graph are needed, but both kinds will be loaded and followed fully to generate the graph.

When generating documentation on the fly, this causes a lot of wasted CPU cycles, potentially loading runtime code that can't impact the typing of the code, and when creating a "runtime" code graph, wastes cycles on analysing type only dependencies.

We should have an option when building the graph to have _All_ _TypeOnly_ and _CodeOnly_ options and visiting only those modules as appropriate.

PR #98, opened 2022-01-07
feat: ability to build type only and code only graphs

kitsonk:
Closes: #95

This should make it easier to generate things like docs for dnt.


dsherret (approved):
Nice! LGTM :)

PR #101, opened 2022-01-12
Make source::Loader easier to use

lucacasonato:
This commit removes the need to explicitly return the passed in
specifier from the returned source::Loader::load function at the cost of
one extra heap allocation.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -1525,7 +1525,7 @@ impl<'a> Builder<'a> {
       let fut = self
         .loader
         .load(&specifier, is_dynamic)
-        .map(move |res| (specifier.clone(), res));
+        .map(move |res| (specifier, res));
       self.pending.push(Box::pin(fut));
     }
   }
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -18,6 +18,8 @@ use deno_ast::MediaType;
 use deno_ast::ParsedSource;
 use futures::stream::FuturesUnordered;
 use futures::stream::StreamExt;
+use futures::Future;
+use futures::FutureExt;
 use serde::ser::SerializeMap;
 use serde::ser::SerializeSeq;
 use serde::ser::SerializeStruct;
@@ -31,6 +33,7 @@ use std::collections::BTreeMap;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::fmt;
+use std::pin::Pin;
 use std::rc::Rc;
 use std::sync::Arc;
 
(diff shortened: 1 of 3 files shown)


bartlomieju (commented):
I'm not sure this is correct - when loading a file you might get a different specifier if there were redirects along the way?

lucacasonato:
> I'm not sure this is correct - when loading a file you might get a different specifier if there were redirects along the way?

Those are returned through the `specifier` property in `source::LoadResponse`

kitsonk:
No, because the specifier in the load response is the "final" specifier, being the redirect, and the outer specifier is the requested specifier to know which request is now complete to remove it from the pending loads.

Making this change is likely to break things.

lucacasonato:
What would it break? The caller of `load` does not need to be returned the original specifier, as it already has the original specifier (as it called `load` with that specifier).

The outer specifier is still there, it is just added by the caller of `load()` now (`deno_graph`), instead of the implementer of `Loader::load`.

kitsonk:
But the results are unordered futures, which are processed in the loop, so while `load` knows, there was no easy way I was aware of to store the requested specifier in relationship to the future:

https://github.com/denoland/deno_graph/blob/a73259d0cf8cdd734305df1bb03a3a1186e2426f/src/graph.rs#L1442-L1445

And the original, requested specifier is needed to compare to the load result, in order to determine if a redirection occurred.  An option could be to add a `maybe_redirected_specifier` to the load result.  Or some sort of "weak map" to associate the `self.pending` future with the requested specifier.

src/graph.rs
```diff
@@ -1515,7 +1521,12 @@ impl<'a> Builder<'a> {
         .graph
         .module_slots
         .insert(specifier.clone(), ModuleSlot::Pending);
-      self.pending.push(self.loader.load(specifier, is_dynamic));
+      let specifier = specifier.clone();
+      let fut = self
+        .loader
+        .load(&specifier, is_dynamic)
+        .map(move |res| (specifier, res));
+      self.pending.push(Box::pin(fut));
```
lucacasonato, line 1529: @kitsonk Please see here. The specifier is still added to the future stored in the `FuturesUnordered`. It is just provided by the inserter (`deno_graph`), instead of the implementer of `Loader`.

lucacasonato: This is what my comment about the extra heap allocation was about :-)

kitsonk: Well, if I actually read the PR versus just the description! 😞  🤦  This is what I get for reading things on my phone late at night.

kitsonk (approved):
LGTM

This is what I had wanted to do originally, but didn't know how.  Thank you... 🙏

PR #104, opened 2022-01-12
feat: add reporting capability

### LINKED ISSUE #103: Add "ProgressReporter" to track module graph building progress
It would be great if the graph builder allowed one to slot in a `ProgressReporter` into the graph builder that would be used to report graph building progress to the caller. How I imagine this working like so:

```rs
struct GraphLoadStats {
  /// The number of unique module specifiers that have been seen by the graph,
  /// that are done loading and parsing.
  completed_modules: usize,
  /// The number of unique module specifiers that have been seen by the graph,
  /// including ones where loading is still "in progress".
  seen_modules: usize,
}

trait ProgressReporter {
  /// This function is called after each module is done loading and parsing.
  fn on_load(&self, specifier: &Url, stats: GraphLoadStats);
}
```

This allows for progress indicators to display the progress of building the graph.

lucacasonato:
This commit adds the "Reporter" trait. Users can pass implementations of
this trait to `create_graph` which will then call the `on_load` method
on the reporter for each module load. This can be used to display a
progress indicator of the graph building.

Reporting is not exposed in the JS API right now.

Closes #103


kitsonk (approved):
LGTM - pending linting issues...

I would be comfortable with "too many args" ignore on the `create_graph`...

devingfx:
This looks great!
Any plan to expose that in the JS API?

PR #108, opened 2022-01-20
refactor: Add CreateGraphOptions, BuilderOptions

bartlomieju:
Rewrite "create_graph", "create_code_graph" and "create_type_graph"
to have fewer arguments. "CreateGraphOptions" struct was added, that
has a "Default" trait implementation.

---- 

I find it quite hard and confusing to understand what's going on
because of `create_graph` having 8 arguments, while most of them are `None`.

Ideally we could have shorthand method for `CreateGraphOptions` like:
```
CreateGraphOptions {
   roots: vec![spec.to_string()],
   loader: &mut some_loader,
   ..Default::default()
}
```
however, because of `roots` and `loader` being required it's not possible to derive `Default` trait. Maybe we could use `CreateGraphOptions::default(roots, loader)` method?


nayeemrmn:
> however, because of `roots` and `loader` being required it's not possible to derive `Default` trait. Maybe we could use `CreateGraphOptions::default(roots, loader)` method?

I think normally those would be excluded from the options struct and kept as positional args.

bartlomieju:
@nayeemrmn thanks for suggestion, totally obvious but I missed it. I think this is a lot cleaner.

src/lib.rs
```diff
@@ -49,60 +49,56 @@ cfg_if! {
...
-      maybe_parser: Option<&dyn SourceParser>,
-      maybe_reporter: Option<&dyn Reporter>,
+      loader: &'_ mut dyn Loader,
+      options: CreateGraphOptions<'_>,
+      build_kind: BuildKind,
     ) -> ModuleGraph {
       let default_parser = ast::DefaultSourceParser::new();
-      let source_parser = maybe_parser.unwrap_or(&default_parser);
+      let source_parser = options.maybe_parser.unwrap_or(&default_parser);
       let builder = Builder::new(
```
bartlomieju, line 70: It seems `Builder` could use similar cleanup

src/lib.rs
```diff
@@ -49,60 +49,56 @@ cfg_if! {
...
+    pub struct CreateGraphOptions<'a> {
+      pub is_dynamic: bool,
+      pub maybe_imports: Option<Vec<(ModuleSpecifier, Vec<String>)>>,
+      pub maybe_resolver: Option<&'a dyn Resolver>,
+      pub maybe_locker: Option<Rc<RefCell<Box<dyn Locker>>>>,
+      pub maybe_parser: Option<&'a dyn SourceParser>,
+      pub maybe_reporter: Option<&'a dyn Reporter>,
+    }
+
+    async fn create_graph_inner(
```
kitsonk, line 62: This is confusing and feels unnecessary, if we are refactoring, let's refactor.

kitsonk (changes requested):
You need to rebuild the WASM and ensure whatever impacts on the JS API are handled properly.

bartlomieju:
> You need to rebuild the WASM and ensure whatever impacts on the JS API are handled properly.

Rebuilt WASM, it seems JS APIs are not impacted by this change (because it uses `Builder` directly)

src/lib.rs
nayeemrmn: I assume the suggestion is to just have one `create_graph()` with the `BuildKind` in options since we're breaking anyway, or at least I think we should do that.

kitsonk: Yes, as well as refactor the builder to use an "options bag" as well.  Might as well go for it.

dsherret (commented):
Reviewed, but waiting on changes (GitHub forces me to leave a comment)

It's nice to convert these to options structs.

kitsonk (commented):
I know it is more, but I am wondering about maybe merging the roots with the build kind, so we only have one positional argument and we can drop the module kind with the build type again.  Something like this:

```rust
enum BuildKind {
  All(Vec<(ModuleSpecifier, ModuleKind)>),
  CodeOnly(Vec<(ModuleSpecifier, ModuleKind)>),
  TypeOnly(Vec<ModuleSpecifier>),
}
```

bartlomieju:
Sure, I'll do that.

bartlomieju:
@kitsonk do you have any objections to moving arguments of `Builder::build()` into `Builder::new()`?

So instead of:
```rust
let builder = Builder::new(
  roots,
  loader,
  source_parser,
  BuilderOptions {
    is_dynamic_root: options.is_dynamic,
    maybe_resolver: options.maybe_resolver,
    maybe_locker: options.maybe_locker,
    maybe_reporter: options.maybe_reporter,
  }
);
builder.build(build_kind, options.maybe_imports).await
```
we would do:
```rust
let builder = Builder::new(
  build_kind,
  loader,
  source_parser,
  BuilderOptions {
    is_dynamic_root: options.is_dynamic,
    maybe_resolver: options.maybe_resolver,
    maybe_locker: options.maybe_locker,
    maybe_reporter: options.maybe_reporter,
    maybe_imports: options.maybe_imports
  }
);
builder.build().await
```

kitsonk:
> @kitsonk do you have any objections to moving arguments of `Builder::build()` into `Builder::new()`?

There was some reason at one point, no specific objection, but it maybe worth taking a look at how it is consumed in CLI, as IIRC there were some challenges in the flow there that lead to that separation of options, but it maybe just me being crazy.

dsherret:
I _think_ we can close this one because it's already done in a separate PR, but please re-open if that's not the case.

PR #117, opened 2022-01-25
feat: improve resolver interface

kitsonk:
This goes with a single enum response, for a resolver:

```rust
pub enum ResolveResponse {
  Amd(ModuleSpecifier),
  CommonJs(ModuleSpecifier),
  Err(Error),
  Esm(ModuleSpecifier),
  Script(ModuleSpecifier),
  Specifier(ModuleSpecifier),
  SystemJs(ModuleSpecifier),
  Umd(ModuleSpecifier),
}
```


src/source.rs
```diff
@@ -89,10 +88,40 @@ pub trait Locker: fmt::Debug {
...
-#[derive(Debug, Clone, PartialEq)]
-pub struct ResolveResult {
-  pub specifier: ModuleSpecifier,
-  pub kind: ModuleKind,
+#[derive(Debug)]
+pub enum ResolveResponse {
+  Amd(ModuleSpecifier),
+  CommonJs(ModuleSpecifier),
+  Err(Error),
+  Esm(ModuleSpecifier),
```
lucacasonato, line 96: Why is `Esm` separate from `Specifier`?

kitsonk: `Specifier` indicates the resolver isn't really doing anything except resolving the specifier, where as `Esm` is asserting it is specifically an ESM module.  Current behaviour is that `Specifier` always gets assigned `ModuleKind::Esm` but we may want to change that in the future, or potentially perform further analysis on the module to determine the module kind.

lucacasonato: Ok, this still confuses me, but let's do it.

bartlomieju: @kitsonk could you add this info as doc comments?

bartlomieju (approved):
LGTM

dsherret (approved):
LGTM too

PR #118, opened 2022-01-25
fix: only analyze JSDoc imports in JavaScript modules

kitsonk:
This skips JSDoc type only imports on non-JavaScript only files.

This isn't a complete fix for denoland/deno#13486 as it should still be a warning on a type only dependency, as I am sure there is other JavaScript code out there with TypeScript imports, but extension-less and we don't want to fail on those.


bartlomieju (approved):
LGTM

PR #120, opened 2022-01-26
feat: add ResolveResponse::to_result()

kitsonk:
Found a pattern that should be in deno_graph when refactoring.


bartlomieju (approved):
LGTM

src/source.rs
```diff
@@ -112,6 +112,21 @@ pub enum ResolveResponse {
   Umd(ModuleSpecifier),
 }
 
+impl ResolveResponse {
+  pub fn to_result(self) -> Result<ModuleSpecifier, Error> {
```
dsherret, line 116: Maybe `to_specifier` instead?

dsherret (approved):

src/source.rs
kitsonk: Hmmm... I would prefer result, because it unwraps the enum to a result.

dsherret: What about `try_specifier()` or something like that?

PR #124, opened 2022-01-29
feat: expose ParsedSource in WASM Module

littledivy:


kitsonk:
What is the use case?

Exposing this is a really complex structure.  I am not sure we want to do it. If we can understand the use case, we should expose a more usable structure in JavaScript.

littledivy:
No longer needed for eszip. Closing.

PR #126, opened 2022-01-31
chore: use wasmbuild instead of _build.ts

littledivy:


README.md
```diff
@@ -215,27 +215,16 @@ be added as follows:
 ```
 > rustup target add wasm32-unknown-unknown
 > cargo install -f wasm-bindgen-cli
+> deno install --unstable -A -f -n wasmbuild https://raw.githubusercontent.com/denoland/wasmbuild/d51c673a321ecf4f336a467bb1a4a5327e783481/main.ts
```
kitsonk, line 218: Can we be more precise than `-A`?

kitsonk: Also, can we publish on `deno.land/x`?

kitsonk (commented):
Was the lib rebuilt with the command and build steps? There seems to be no changes to the `/lib` which I find surprising.

README.md
```diff
@@ -215,27 +215,16 @@ be added as follows:
...
-
-```
-> ./_build.ts
-```
-
-But can be manually invoked like:
-
 ```
-> deno run --unstable _build.ts
+> wasmbuild --no-default-features --features wasm
```
dsherret, line 225: Until wasmbuild stabilizes we should probably continue using a script with a pegged version (ex. getting https://github.com/denoland/wasmbuild/issues/3 done)

kitsonk:
I really think we need to figure out a solution that works without using `deno install`, or at least if it is `deno install`.  Right now, on OSX and Linux systems, once setup you just do `./_build.ts` and changes to the build or the build dependencies are reflected in the build.

We should also take this opportunity to have CI do the building, but there is still a need to do a local build for testing and validation.

dsherret:
This is resolved in main now.

PR #130, opened 2022-02-04
feat: graph supports external and built in modules

### LINKED ISSUE #127: Allow representation of external and built-in dependencies
The graph should allow a loader to represent a load request with an `External` and `BuiltIn` load result.  These would become the `ModuleKind` for the module slot and no dependency analysis would be performed.  This would be an indictor to implementors that these require special loading at runtime.

kitsonk:
Closes #127

I think it is functionally complete, and I have all the existing tests running.  I just need to add tests for the new stuff,

But basically it is a minor refactor for existing code which goes from:

```rust
LoadResponse {
  specifier,
  maybe_headers,
  content
}
```

To:

```rust
LoadResponse::Module {
  specifier,
  maybe_headers,
  content
}
```

And the new external and built-in would be:

```rust
LoadResponse::External(specifier);
LoadResponse::BuiltIn(specifier);
```

They will be represented in the graph as:

```rust
ModuleKind::External;
ModuleKind::BuiltIn;
```


littledivy (approved):
API wise LGTM! I don't think for eszips we need even external specifiers slots in the graph (?) but makes sense for other usecases.

kitsonk:
> I don't think for eszips we need even external specifiers slots in the graph

`deno compile`'s usage of eszip may require it

PR #137, opened 2022-02-13
fix: coalesce multiple same-specifier imports correctly

### LINKED ISSUE #136: Static and dynamic import of the same module should be static dep
```ts
// temp.ts
import "./nonexistent.ts";
await import("./nonexistent.ts");
```
`deno cache temp.ts` gives no error, since the last import was dynamic and overrides `Dependency::is_dynamic`.

nayeemrmn:
Fixes #136. Also generally makes the first occurrence (rather than the last) take priority for duplicate imports.


dsherret (approved):
LGTM. Thanks @nayeemrmn!

PR #149, opened 2022-04-27
feat: add source maps to graph

### LINKED ISSUE #146: Identify source map URLs in modules
When analyzing sources, any "upstream" source maps should be identified and become part of the module graph.  This would allow consumers of the graph information to be able to easily identify any upstream source maps and consume them.

This would help enable denoland/deno#4499

kitsonk:
Closes #146

This adds source maps to the module graph.  When adding a module to the graph, it will be checked to see if it references a source map, and if it does, it will add it to the graph.  If it is an inline source map, it will be decoded and added as a JSON value.  If it is just a reference to a file, the parsed specifier it references will be added to the graph.

This will enable the CLI to be able to utilise upstream source maps when running code and re-mapping code points.


bartlomieju (approved):
LGTM, no comments

src/source_map.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum ParsedSourceMap {
+  Value(Value),
+  Url(ModuleSpecifier),
+}
+
+fn find_source_map_url(parsed_source: &ParsedSource) -> Option<String> {
+  for comment in parsed_source.comments().trailing_map().values() {
```
dsherret, line 25: I think this will be in an unstable order so if there are multiple of these source map url comments it may match one or the other. If we're only looking for the trailing comments in a file, then think this would work: `parsed_source.comments().get_trailing(parsed_source.program.span().hi)` (but I forget how swc stores the trailing comments... maybe it's on the last token?)

src/source_map.rs
```diff
@@ -0,0 +1,125 @@
...
+) -> Option<ParsedSourceMap> {
+  let source_map_url = find_source_map_url(parsed_source)?;
+  let source_map_specifier =
+    if let Ok(url) = ModuleSpecifier::parse(&source_map_url) {
+      Some(url)
+    } else if let Ok(url) = specifier.join(&source_map_url) {
+      Some(url)
+    } else {
+      None
+    }?;
```
dsherret, line 51: I think this could maybe use `or_else` here:

```rs
let source_map_specifier = ModuleSpecifier::parse(&source_map_url)
  .or_else(|| specifier.join(&source_map_url))
  .ok()?;
```

dsherret (approved):
LGTM. Just one small potential issue.

PR #151, opened 2022-04-28
[WIP] feat: add web assembly modules to graph

kitsonk:
This commit adds support for WebAssembly modules in the graph. Imports
inside of the WebAssembly are parsed out using `wasmparser`, and are
further analyzed by the graph like other dependencies.

Some follow up work needs to be done to represent source content as
`Uint8Array` in JS.


kitsonk:
Ah, I see that you have the other PR that converts everything to `[u8]` including the JS side...  We can revert the deserialization once we do that, though it would be great to allow both a string or array buffer on the JavaScript side, but 🤷

kitsonk:
Stale and would need to be reworked.

Issue #132, opened 2022-02-08
Support multiple ranges for an import in the graph


kitsonk:
Currently the graph cannot represent multiple locations an import specifier might be imported in a file.  For example, an import can be like this:

```ts
import * as a from "https://example.com/a";
import { default as b } from "https://example.com/a";
```

In addition, in theory, a module can be statically and dynamically imported:

```ts
import * as a from "https://example.com/a";
const b = await import("https://example.com/a";
```

Or in TypeScript the same specifier can be provided for a type only import:

```ts
import { a } from "https://example.com/a";
import type { A } from "https://example.com/a";
```

Currently we only store the last range for the specifier in the graph.  This makes it difficult when trying to rewrite specifiers (like in dnt or `deno vendor`) as well as means that diagnostics are only issued for the "last" import location, not all import locations for a given specifier.

We should enhance the graph to be able to store this information, which is likely going to be a breaking change, but one that increases the usability of the graph.

nayeemrmn:
This is a must for proper import assertion handling. Import assertion errors should be stored under each of these, not occupy module slots.

dsherret:
I've run into needing this for `deno vendor` again while working on the import map only solution. It also needs to know the text for the `// @deno-types="..."` and `/// <reference types="..." />` directives so that it can tell if it needs to map those.

nayeemrmn:
Closed in #251.

PR #155, opened 2022-05-20
fix: error on conflicting assertions

lucacasonato:
Previously, if a specifier was imported more than once, with conflicting
assertions, we'd only really consider the final import. This is wrong:
conflicting assertions are an error in-themselves already. I think this
behaviour is still wrong, because if a module is imported with incorrect
assertions later in the graph after the module has already been loaded by
an import with a correct assertion, no error will be raised.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -133,6 +133,7 @@ pub enum ModuleGraphError {
     actual_media_type: MediaType,
     expected_media_type: MediaType,
   },
+  ConflictingAssertions(ModuleSpecifier),
   LoadingErr(ModuleSpecifier, Arc<anyhow::Error>),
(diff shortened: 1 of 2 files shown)


nayeemrmn:
Long term we should do this with #132 instead, and associate import assertion errors with each import. With per-import errors being a third kind on top of resolution errors and module slot errors. Maybe this patches some error cases, but the data modelling wrt import assertions is still in a very bad state.

bartlomieju:
Closing as stale — import assertions have since been replaced by the import attributes model, and the conflicting-import handling has been redesigned (along the lines of the per-import error modeling suggested in the thread). The code path this targets no longer exists, and the branch has long-standing conflicts. Thanks regardless!

PR #164, opened 2022-07-15
Implement `std::Error::source()` for errors

honzasp:
`ModuleGraphError` and `ResolutionError` did not implement `Error::source()`, so if a user-provided `Resolver` or `Loader` returns an error, its cause chain is lost.

This PR implements `source()`, but it keeps the existing implementation of `Display`. Unfortunately, this means that when anyhow prints the cause chain of errors, it might print some error messages multiple times. For example, when we print a `ModuleGraphError::LoadingErr(err)`, then `anyhow` first prints the `ModuleGraphError`, which displays `err`. `anyhow` then prints the source of the `ModuleGraphError`, which is `err`, so `err` is displayed again!

There are two possible solutions to this problem:

1. Change implementation of `Display`, so that `ModuleGraphError::LoadingErr(err)` displays a fixed error message (`"The module could not be loaded"`). The disadvantage is that if the caller does not also print the cause chain, the user will not see the actual error!
2. Change implementation of `Error::source()`, so that `ModuleGraphError::LoadingErr(err)` returns `err.source()` instead of `err`. This means that `err` will not be displayed twice, but `err` will no longer be present in the cause chain.


kitsonk (approved):
LGTM

PR #168, opened 2022-08-01
BREAKING CHANGE - fix: move imports out of modules

### LINKED ISSUE #166: Can not import JSON config file if the config file specifies jsxImportSource
```js
// main.ts
import config from "./deno.json" assert { type: "json" };
const obj = { config };
export default obj;
```

```json
// deno.json
{
  "compilerOptions": {
    "jsx": "react-jsx",
    "jsxImportSource": "preact"
  },
  "importMap": "./import_map.json"
}
```

```json
// import_map.json
{
  "imports": {
    "preact/": "https://esm.sh/preact@10.8.2/"
  }
}
```

```
$ deno run -A main.ts
error: Loading unprepared module: file:///tmp/jsx_config_issue/deno.json
```

I think what is happening is that the config file is stored in the module graph as some sort of special entry, which is causing this weird behaviour.

kitsonk:
Fixes #166

This removes the module kind synthetic, and instead, dependencies that are injected into the graph (like "types" or a JSX runtime) are added instead as graph imports. This specifically addresses the issue where external imports to the graph from `deno.json` config files made it impossible for the `deno.json` to also be imported as a JSON module.

This is breaking because it removes the module kind `Synthetic` (as well as adds a new optional property on the graph called `imports`).  This should have minimal impact on consumers though, as `Synthetic` was generally an internal structure and was filtered out and not treated as a "real" module.


lucacasonato (commented):
This is a fine solution, but I think it has overlap with `roots`. What if roots was a `([specifier: string, base: string] | string)[]`?

kitsonk:
> This is a fine solution, but I think it has overlap with `roots`. What if `roots` was a `([specifier: string, base: string] | string)[]`?

We have had the `build(roots, maybe_imports)` signature for a long time.  `roots` are clearly modules that exist as the starting point for dependency analysis. Where as imports are modules which have a non-analyzeable base.  Previously these imports though existed as a synthetic module.  Now they exist as a graph import.  Overloading the roots so that it is maybe a module or maybe an meta data import doesn't improve things in my mind, and doesn't mean that they still need different representations in the graph.

Imports are ambient items, like library type definitions or JSX runtimes that are not imported dependencies of the roots or any other part of the graph.

lucacasonato:
Mh, I guess I have always just thought of `roots` as specifiers to start analysis from. I guess my thinking is that `types` and `jsxImportSource` are also just specifiers to start analysis from. The primary difference being that right now roots must be fully qualified specifiers, whereas to support the other usecases we need a way to specify unresolved specifiers as an entrypoint. The idea behind `[specifier: string, base: string]` is that you provide the parts necessary to specify the parts required to create the fully qualified specifier that the graph can then resolve into a fully qualified specifier using the resolver.

lucacasonato:
I'll let David weigh in and approve. I don't have a very strong opinion on this either way.

dsherret (approved):
LGTM! I don't have a strong opinion.

The word "imports" seems like it could be more specific, but it's been already used and I can't think of anything better.

kitsonk:
I have thought about it a bit more. I think both abstractions are valid. Had we introduced ambient imports as part of the roots, then clearly it would make sense to continue down that path.  I still think there is a distinction between the root of the graph and these ambient imports which need to be resolved.  It is akin to `files` and `compilerOptions.types` in `tsc`.

If we did change it, it would be a significant breaking change, with every integration breaking, as well as making a breaking change in the output of a graph too.  This current solution is only affecting situations where there was a dependency on the "synthetic module" in the output, which should be close to 0.

Given that there isn't a significant upside or downside with either approach, I would prefer to go with this less breaking approach.

PR #176, opened 2022-08-23
fix: native default `jsxImportSource` resolution

lucacasonato:
This commit adds support for natively resolving a default
jsxImportSource specified in a config file per module.

The fix that this allows for upstream is that all
`jsxImportSource`s should be resolved relative to the module
at all times. Right now we resolve relative to the config
file in the graph, and relative to the module at runtime.

Towards denoland/deno#15559

Upstream integration: https://github.com/denoland/deno/pull/15561


kitsonk (approved):
LGTM and this all makes sense.

PR #183, opened 2022-10-03
refactor: Change public API

bartlomieju:
Trying to get rid of some code smells.

- Added "GraphOptions" that stores all the optional fields
- changed "create_graph", "create_code_graph" and "create_type_graph"
to accept "GraphOptions" instead of 8-9 arguments.

In a follow up I'd like to rework `Builder` and make it public as well.


dsherret (approved):
LGTM. Nice cleanup!

kitsonk (approved):
LGTM... yeah our options were getting out of hand...

PR #184, opened 2022-10-05
refactor: consolidate Amd, CommonJs, SystemJs, and Umd modules into Script modules

bartlomieju:
With removal of "compat mode" in the CLI there is no use for
CommonJS modules in graph, discussed offline with David
and we feel this should be dropped.


src/source.rs
```diff
@@ -106,8 +106,6 @@ pub trait Locker: fmt::Debug {
 pub enum ResolveResponse {
   /// The resolved specifier where the module is an AMD module.
   Amd(ModuleSpecifier),
```
dsherret, line 108: @kitsonk do you think it's worth still having Amd, SystemJs, Umd?

dsherret (approved):
LGTM

src/source.rs
kitsonk: 🤷 I still think it is worth having it support CommonJS, as it makes it more general purpose.  Just because Deno CLI doesn't use it anymore doesn't make it useful.

kitsonk:
As I stated in the reply to a comment, I am not supportive of this PR. Yes, while we added support for resolvers to mark modules as CommonJS to support Deno CLI, we also made deno_graph more general purpose. Deno CLI isn't the only user of deno_graph. It has 5,500 MAU currently on the JavaScript side let alone however many using the Rust crate.

Outside of Deno CLI no longer using the feature is leading you to want to diminish the feature set of deno_graph?

dsherret:
The main motiviation to removing it is less code to maintain and it simplifies deno graph a little bit. We're not using it and it doesn't look like anyone is using this. I don't have strong opinions though.

https://cs.github.com/?q=ResolveResponse%3A%3ACommonJs&p=2&pt=b0bdd64327cd06031ac12ca1d3663982eec707a3f19132c832dce8e0a314b55dd2b45ba0bb51b6f108652ac82af4f2302a5da4fe7fdea124c64ba6138156aa3ed4173ee22021d434d90b3f955fc6b511d8aee1d0fe6d9b8f1db52d25e06c3f774d9a4a77f01d0bc27154693d1606975d8df68d152c454c0e77d0fe47357a95deca5aa3f07d29cd89&scope=&scopeName=All+repos

dsherret:
We can add this back later, but it doesn't look like anyone is using this and IMO it's questionable this is useful in its current form, so let's simply and combine to just `Script` for the time being.

PR #192, opened 2022-10-31
Remove support for Locker trait

### LINKED ISSUE #189: Suggestion: Remove Locker from deno_graph
It looks like we could remove the concept of "locking" from deno graph and instead keep this functionality completely in the CLI. All we really need is a list of modules and their sources to do this, so this could help us simplify deno graph and we could also do this work more easily in parallel in the CLI.

bartlomieju:
This logic is very dependent on the state of `deno` CLI and after numerous discussions with David we deemed it's better to keep it completely in the CLI.

Closes #189


dsherret (approved):
LGTM. Someone can operate on the files themselves to do lockfile behaviour.

PR #212, opened 2023-02-04
fix: trim quotes of import/export specifiers

relrelb:
swc provides a span over the import/export specifier including the enclosing quotes. we want the just specifier without the quotes.

For https://github.com/denoland/deno/issues/17645.


src/ast.rs
```diff
@@ -298,6 +298,12 @@ impl ParsedSourceStore for CapturingModuleAnalyzer {
 fn analyze_dependencies(
   parsed_source: &ParsedSource,
 ) -> Vec<DependencyDescriptor> {
+  fn trim_quotes(range: SourceRange) -> SourceRange {
+    // swc provides a span over the import/export specifier including the
+    // enclosing quotes. we want the just specifier without the quotes.
+    SourceRange::new(range.start + 1, range.end - 1)
```
dsherret, line 304:
```
cargo test
```

...will show some failures that need to be updated (let me know if you want help with it). Once all the tests are fixed, the next thing to do would be to run:

```
deno task build
```

And finally:

```
deno test -A
```

relrelb: Done.

dsherret (approved):
Thanks @relrelb! There's a lot of other changes in deno_graph that are still going to be done, so I'll upgrade it in the CLI sometime in the next few days.

PR #228, opened 2023-02-12
refactor: check assertions per import statement

### LINKED ISSUE #132: Support multiple ranges for an import in the graph

### LINKED ISSUE #160: Don't enforce assertions on type-only imports
There exists `TS2822 [ERROR]: Import assertions cannot be used with type-only imports or exports.`, so we shouldn't check assertions for type-only imports. https://github.com/denoland/deno/issues/14913

nayeemrmn:
Closes #132
Fixes #160
Supersedes #155

Assertions appropriately no longer influence module loading/parsing. Instead they are checked in a pass at the end; resulting errors are stored in `Import::errors: Vec<ImportError>`. 'Import errors' are a third kind of error alongside module slot errors and resolution errors which are now checked in `ModuleEntryIterator::validate()`.


src/analyzer.rs
```diff
@@ -182,6 +182,44 @@ impl ImportAssertions {
       _ => None,
     }
   }
+
+  pub fn is_known_none(&self, key: &str) -> bool {
+    match self {
+      ImportAssertions::None => true,
+      ImportAssertions::Known(map) => !map.contains_key(key),
+      _ => false,
```
dsherret, line 190: Can you change `_` here to be explicit about the variant? That will make the code easier to read without knowing all the variants and cause compile errors here when the variants change. I know we don't do that everywhere in the code here, but we should do it more often.

src/analyzer.rs
```diff
@@ -182,6 +182,44 @@ impl ImportAssertions {
       _ => None,
     }
   }
+
+  pub fn is_known_none(&self, key: &str) -> bool {
```
dsherret, line 186: I'm confused by this method name because it does `!map.contains_key`. Maybe this should be `has(&self, key: &str)` then we can invert everything and do `!has(` when using it?

src/analyzer.rs
```diff
@@ -182,6 +182,44 @@ impl ImportAssertions {
...
+        for (key, assertion) in map {
+          match assertion {
+            ImportAssertion::Known(value) => {
+              assertions.insert(key.clone(), value.clone());
+            }
+            ImportAssertion::Unknown => return None,
+          }
+        }
+      }
+      _ => return None,
```
dsherret, line 213: We should convert this wildcard to be explicit.

src/graph.rs
```diff
@@ -1735,16 +1817,7 @@ impl<'a, 'graph> Builder<'a, 'graph> {
...
-          let assert_types =
-            self.pending_specifiers.remove(&specifier).unwrap();
-          for maybe_assert_type in assert_types {
-            self.visit(
-              &specifier,
-              &response,
-              maybe_assert_type,
-              maybe_referrer.clone(),
-            )
-          }
```
nayeemrmn, line 1747: This is the current method for handling different assertions being made against the same module... re-parse it with each one (because that's where assertion errors are generated). And it seems like there's no uniqueness in the assert types in `pending_specifiers` because it includes the range... so are we re-parsing the module for every import that was found for it until it was visited?

src/analyzer.rs
```diff
@@ -182,6 +182,44 @@ impl ImportAssertions {
...
+      _ => false,
+    }
+  }
+
+  pub fn is_empty(&self) -> bool {
+    matches!(self, ImportAssertions::None)
+  }
+
+  /// Try to get a flat dictionary, successful if there are no unknowns.
+  pub fn maybe_static(&self) -> Option<HashMap<String, String>> {
```
dsherret, line 199: Where is this used? Is it necessary? We should minimize having methods on the that allocate while analyzing a graph. For example, maybe this should instead be a method that verifies the `ImportAssertions` then someone could use `.get` afterwards instead or this should return a new struct where this method verifies the `ImportAssertions` then has a reference to the old map while providing some `.get` methods.

src/analyzer.rs
```diff
@@ -182,6 +182,44 @@ impl ImportAssertions {
...
+    let mut assertions = HashMap::new();
```
dsherret, line 200: If really necessary, move within the match and specify the capacity based on the existing map:

```suggestion
    let mut assertions = HashMap::with_capacity(map.len());
```

src/analyzer.rs
```diff
@@ -182,6 +182,44 @@ impl ImportAssertions {
...
+    match self {
+      ImportAssertions::None => {}
+      ImportAssertions::Known(map) => {
+        for (key, assertion) in map {
+          match assertion {
+            ImportAssertion::Known(value) => {
+              assertions.insert(key.clone(), value.clone());
+            }
+            ImportAssertion::Unknown => return None,
```
dsherret, line 209: We do we return `None` for everything if a specific import assertion is not known? Maybe this method seems too specific and shouldn't exist in deno_graph?

src/graph.rs
```diff
@@ -130,37 +140,28 @@ impl Range {
...
+  // Note: Resolution and import errors are special in that they don't occupy
+  // module slots, whereas the other variants do.
+  ResolutionError(ResolutionError),
+  ImportError(ImportError),
 }
 
 impl ModuleGraphError {
   pub fn specifier(&self) -> &ModuleSpecifier {
     match self {
+      Self::ImportError(_) => panic!("Import errors don't have specifiers"),
```
dsherret, line 160: We should remove this panic. Either all the errors should have a specifier or this should become `Option<&ModuleSpecifier>`. I'm not sure what's changed that there wouldn't be a specifier anymore though?

dsherret (commented):
I'll review the rest of it later.

src/graph.rs
```diff
@@ -1655,18 +1731,25 @@ pub(crate) struct AssertTypeWithRange {
...
   resolver: Option<&'a dyn Resolver>,
   pending: FuturesUnordered<LoadWithSpecifierFuture>,
-  pending_specifiers:
-    HashMap<ModuleSpecifier, HashSet<Option<AssertTypeWithRange>>>,
-  dynamic_branches:
-    HashMap<ModuleSpecifier, (Range, Option<AssertTypeWithRange>)>,
+  dynamic_branches: HashMap<ModuleSpecifier, Range>,
   module_analyzer: &'a dyn ModuleAnalyzer,
   reporter: Option<&'a dyn Reporter>,
+  module_metadata: HashMap<ModuleSpecifier, ModuleMetadata>,
```
dsherret, line 1752: I'm not sure we need this because this information should be avaiable straight from the graph? I think this will maybe fail if someone calls `build` multiple times.

src/analyzer.rs
nayeemrmn: Ah I was going to use this as an alternative serialization for static imports but changed my mind (deferred for the future?), I'll remove this method.

src/graph.rs
nayeemrmn: > I'm not sure what's changed that there wouldn't be a specifier anymore though?

Yeah I've been meaning to complain about this method... the specifier returned by `ResolutionError` has a drastically different meaning than the other variants (referrer vs target). Similar with `maybe_range()`. `ImportError` kind of aligns with `ResolutionError` on that.

It's a symptom of `ModuleGraphError` mixing errors that can occupy module slots vs errors that represent arrows between them. A year ago on a previous attempt at this fix I tried separating them more thoroughly but the boilerplate was extreme. So yeah I think we should just make it return an `Option`.

src/graph.rs
nayeemrmn: It's to work around borrow issues at https://github.com/denoland/deno_graph/blob/af1009eb0e5cd9dcb1c851911f48268ed187f7b5/src/graph.rs#L1877-L1885
If there were a workaround limited to `Builder::fill()`, it would probably be to build up the same map but as its own pass just above this.

src/analyzer.rs
nayeemrmn: Then `has()` would have to return true if the presence of associated value of that key is unknown. The method returns true if the given key is 100% known to be absent (that's where we treat it the assertion as `type: "js"`), and returns false if the set of keys is dynamic (can't check assertions) or the key is set to a value (we will use `get()` to check assertions).

src/graph.rs
dsherret: Yeah, we'll have to make that work based on the current module graph. It would be better if we didn't create two sources of truth and just derived from the graph.

src/analyzer.rs
dsherret: True. Maybe add a documentation comment on this?

src/graph.rs
```diff
@@ -130,38 +139,29 @@ impl Range {
...
-  UnsupportedImportAssertionType {
-    specifier: ModuleSpecifier,
-    range: Range,
-    kind: String,
-  },
   UnsupportedMediaType {
     specifier: ModuleSpecifier,
     media_type: MediaType,
     maybe_referrer: Option<Range>,
   },
```
dsherret, line 149: RE https://github.com/denoland/deno_graph/pull/228#discussion_r1103865391 -- I wonder if we should have `ModuleError(ModuleError)` then put `LoadingErr`, `Missing`, `ParseError`, and `UnsupportedMediaType` in there. Then we could move `pub fn specifier(&self) -> &ModuleSpecifier` onto that and then move `maybe_range` into the other errors. What do you think?

src/graph.rs
```diff
@@ -502,25 +497,64 @@ fn is_false(v: &bool) -> bool {
...
-    serialize_with = "serialize_resolution"
-  )]
   pub maybe_type: Resolution,
-  #[serde(skip_serializing_if = "is_false")]
   pub is_dynamic: bool,
-  #[serde(rename = "assertType", skip_serializing_if = "Option::is_none")]
-  pub maybe_assert_type: Option<String>,
+  /// List of code imports which refer to this dependency. Only ones which can
+  /// populate `Dependency::maybe_code`.
+  pub imports: Vec<Import>,
```
dsherret, line 557: I'm not sure if "imports" is the right word here because currently in this PR this could be imports, exports, or typescript references. Perhaps it should be `specifiers` instead?

src/graph.rs
```diff
@@ -1616,6 +1643,54 @@ fn is_untyped(media_type: &MediaType) -> bool {
   )
 }
 
+/// Check import assertions.
+fn check_import_assertions(
+  import: &Import,
+  maybe_module_metadata: Option<&ModuleMetadata>,
+) -> Vec<ImportError> {
+  // Skip assertion checks for dynamic imports for now.
```
dsherret, line 1651: Is this done later? Why aren't dynamic imports checked? Maybe include the answer to "why?" in the comment here.

dsherret (changes requested):
Thanks for this @nayeemrmn. It's looking better than main, but I'm not sure about the name "imports". Perhaps this should be something more generic because it also includes exports and typescript path references?

src/graph.rs
nayeemrmn: The way we discover dynamic module slot errors in `deno_graph` and defer elevating them until runtime, there's no good mechanism for doing that currently in CLI for resolution/import errors so we just leave it for `deno_core`. That and there was a test for it (`test_build_graph_dynamic_json_ignores_assert`) so I just excluded it for simplicity.

I think I'll move this check to `validate()` actually, it's more consistent with module slot errors.

dsherret: Good idea. That sounds a lot better.

src/graph.rs
nayeemrmn: Yeah that's what I did in the previous attempt from last year I mentioned https://github.com/nayeemrmn/deno_graph/blob/786128a3f6330b1bdf7c6408e2ca53eb9c4688e9/src/graph.rs#L297-L302. I'll try it again, it should go better with this version's solution.

test.ts
```diff
@@ -282,7 +297,28 @@ Deno.test({
                   },
                 },
               },
-              "assertionType": "json",
+              "imports": [
+                {
+                  "assertions": {
+                    "Known": {
```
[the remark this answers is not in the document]
nayeemrmn: Yeah, I guess we can switch to blessed fields for known assertions on the serialization layer. Just now wondering if we should store it like that in the first place..

src/graph.rs
dsherret: Ok, imports doesn't sound too bad. You're right that it's more clear.

test.ts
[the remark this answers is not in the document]
dsherret: If everything still works then storing it like that might be better.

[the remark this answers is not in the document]
nayeemrmn: I've been mulling over schemas that:
1. Transfer well to dynamic imports with potential unknowns
2. Don't then add a ton of boilerplate for static imports
3. Don't conflate dynamic/unknown assertions with unspecified ones

And landed on:
```ts
interface Import {
  assertions: "unknown" | Record<string, string>,
}
```
where for static imports you can just pretend the `"unknown" |` isn't there. This would use the `maybe_static()` method from https://github.com/denoland/deno_graph/pull/228#discussion_r1103861394 and just mark dynamic imports as `"unknown"` if there is _any_ dynamic part of it. We have to give up the blessed fields like `assertionType?: string` since it would violate either (1) or (3) but `assertions.type` is pretty much the same.

src/graph.rs
```diff
@@ -848,6 +882,13 @@ impl<'a> ModuleEntryIterator<'a> {
                   ));
                 }
               }
+              for import in &dep.imports {
+                if !import.is_dynamic {
```
dsherret, line 886: I think this should be:

```suggestion
                if self.follow_dynamic || !import.is_dynamic {
```

I just realized the above on line 872 is wrong and should also be similar.

nayeemrmn: Note to correct https://github.com/denoland/deno/blob/f917d2e2c10e0a94e564a9016217e7ce27c8bbee/cli/tools/vendor/build.rs#L88 to false

dsherret: I changed that one to `true` because I think we should validate dynamic imports for vendoring.

test.ts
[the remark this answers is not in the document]
dsherret: Sounds good.

src/graph.rs
nayeemrmn: Okay I thought this had been masking errors that would now be exposed but it only applies to resolution errors. I can see why missing modules should be caught in advance for vendoring

dsherret: That said, I guess that would prevent vendoring of the scenario you described yesterday where someone might conditionally load some plugin code or something. Maybe instead it should be warnings.

src/graph.rs
```diff
@@ -559,12 +613,8 @@ pub struct TypesDependency {
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
 pub enum ModuleKind {
-  /// An asserted module. The import location is required to determine what the
-  /// asserted type is as well as a loader/runtime would want to ensure the
-  /// `MediaType` matches the assertion. Dependency analysis does not occur on
-  /// asserted modules.
-  Asserted,
```
nayeemrmn, line 566: I've merged `ModuleKind::Asserted` into `ModuleKind::Esm`. That info is given by `Module::media_type` and assert-worthy module types should fall in naturally for other purposes. The name seems more related to module system not language so it should be fine.

dsherret: I'm doing a large refactor for npm specifiers atm that's going to remove `ModuleKind` and replace it with an `enum Module`. We're not going to land this PR until deno 2.0 so you'll get some conflicts on this on this one, but I don't think it will be too much trouble.

src/graph.rs
```diff
@@ -138,29 +138,66 @@ impl Range {
 }
 
 #[derive(Debug, Clone)]
-pub enum ModuleGraphError {
+pub enum ModuleError {
```
dsherret, line 141: By the way, we can't land this PR until probably next month in preparation for Deno 2.0. If you want to get some of these non-breaking changes into main faster then you might want to consider refactoring these out ot a separate PR and I'll take a look at it to get it merged quickly.

src/graph.rs
dsherret: I think we could represent it as `Module::Json` instead of `Module::Asserted`. I'll update my PR to that.

nayeemrmn:
Regarding 2.0 breakages, seems like the only breakage is `Dependency::assertionType`. Could we keep that as legacy for a month to unblock this PR?

dsherret:
Probably not because merging this would also require more work upgrading the rest of the code in other repos and that would block progress on getting npm specifiers in deno_graph, which is the priority at the moment because we want to get them in a more ideal state for integration with Deploy and other tooling. It's just waiting a couple weeks until early March.

src/graph.rs
```diff
@@ -137,44 +146,71 @@ impl Range {
...
 
 #[derive(Debug, Clone)]
-pub enum ModuleGraphError {
-  InvalidTypeAssertion {
-    specifier: ModuleSpecifier,
-    range: Range,
-    actual_media_type: MediaType,
-    expected_media_type: MediaType,
-  },
+pub enum ModuleError {
```
dsherret, line 149: Would you be able to pull out these error changes and open a separate PR with them (and for anything else in this PR that's non-breaking for the json)? I'm not sure when we'll be able to land this PR and it would be nice to get this change in.

nayeemrmn:
@dsherret I went ahead and preserved `assertionType` in the serialization for legacy. It's a very self-contained, small addition in `serialize_dependencies()`. This can be considered for landing without breaking `deno info` now.

nayeemrmn:
With import assertions becoming import attributes and them being part of the cache key, this PR goes in the wrong direction for the future. Closing.

PR #236, opened 2023-02-16
refactor: cleanup serialize functions

nayeemrmn:
Simplify serialize definitions for  `module_slots`, `graph_imports` and `dependencies`. Derive as much as possible, deduplicate handling between derived field attributes and overrided serialize functions.


nayeemrmn:
@dsherret Please review

src/graph.rs
```diff
@@ -1138,14 +1105,21 @@ impl<'a> Iterator for ModuleGraphErrorIterator<'a> {
...
+#[derive(Debug, Default, Clone, Serialize)]
 pub struct ModuleGraph {
+  #[serde(skip_serializing)]
   graph_kind: GraphKind,
   pub roots: Vec<ModuleSpecifier>,
+  #[serde(rename = "modules")]
+  #[serde(serialize_with = "serialize_module_slots")]
   pub(crate) module_slots: BTreeMap<ModuleSpecifier, ModuleSlot>,
+  #[serde(skip_serializing_if = "BTreeMap::is_empty")]
+  #[serde(serialize_with = "serialize_graph_imports")]
```
bartlomieju, line 1117: Do these need to use two separate attributes? Can't they be combined onto a single attribute call?

nayeemrmn: You can use comma syntax as well but it would just be four lines instead of two and imo better to separate them like imports. Or if you mean putting the skipping logic in the function, I don't think that's possible because it can't control key inclusion

dsherret (approved):
LGTM. Thanks, @nayeemrmn!

PR #246, opened 2023-03-05
fix: make ts reference paths type-only deps

nayeemrmn:
`<reference path="..." />` doesn't create a runtime dependency. Extracted from #228.
Prevents the following module from failing on a no-check graph validation:
```ts
/// <reference path="./nonexistent.d.ts" />
```


dsherret (approved):
Nice catch! Thanks!

PR #250, opened 2023-03-06
refactor: separate `ModuleError` from `ModuleGraphError`

nayeemrmn:
Towards #228


dsherret (approved):
Thanks!

src/graph.rs
```diff
@@ -137,44 +137,93 @@ impl Range {
 }
 
 #[derive(Debug, Clone)]
-pub enum ModuleGraphError {
+pub enum ModuleError {
+  LoadingErr(ModuleSpecifier, Option<Range>, Arc<anyhow::Error>),
```
dsherret, line 141: I think we should eventually change these all to object syntax so it's more understandable.

PR #251, opened 2023-03-06
feat: store import sites on dependencies

nayeemrmn:
Towards #228. Stops short of introducing `ImportError` / checking assertions because that's where it starts needing to be broken in. They're not included in the json yet since the schema needs a _little_ more scrutiny, but the serialization code is there to iterate on and is usable in the rust API.


src/analyzer.rs
```diff
@@ -182,6 +184,47 @@ impl ImportAssertions {
       _ => None,
     }
   }
+
+  /// Check if a assertion key is known to be absent (not set to a value and not
+  /// undeterminable). This is significant because it is the precise case where
+  /// `type: "javascript"` is asserted.
+  pub fn is_known_none(&self, key: &str) -> bool {
```
dsherret, line 191: Not used yet? I think we can remove this for the purposes of this PR?

src/analyzer.rs
```diff
@@ -182,6 +184,47 @@ impl ImportAssertions {
...
+    }
+  }
+
+  pub fn is_empty(&self) -> bool {
+    matches!(self, ImportAssertions::None)
+  }
+
+  /// Try to get a flat dictionary, successful if there are no unknowns.
+  pub fn maybe_static(&self) -> Option<HashMap<String, String>> {
+    let mut assertions = HashMap::new();
```
dsherret, line 205: We could pre-allocate this by doing `HashMap::with_capacity(map.len())`

src/analyzer.rs
```diff
@@ -182,6 +184,47 @@ impl ImportAssertions {
...
+    match self {
+      ImportAssertions::None => {}
+      ImportAssertions::Known(map) => {
+        for (key, assertion) in map {
+          match assertion {
+            ImportAssertion::Known(value) => {
+              assertions.insert(key.clone(), value.clone());
```
dsherret, line 212: We could avoid cloning and return a `HashMap<&str, &str>`

src/graph.rs
```diff
@@ -470,7 +471,68 @@ fn is_false(v: &bool) -> bool {
...
-#[derive(Debug, Default, Clone, Serialize, PartialEq, Eq)]
+fn serialize_import_assertions<S>(
+  assertions: &ImportAssertions,
+  serializer: S,
+) -> Result<S::Ok, S::Error>
+where
+  S: Serializer,
+{
+  if let Some(map) = assertions.maybe_static() {
+    map.serialize(serializer)
```
dsherret, line 482: This doesn't seem exactly right because we're losing resolution. Overall, I think we should maybe remove the serialization for this completely atm since import assertions might change a lot https://github.com/denoland/deno/issues/17944

dsherret (commented):
Looks mostly good to me. @nayeemrmn would you be able to add a basic test for each kind of import?

src/graph.rs
dsherret, line 482: Serialization here is unstable because the underlying structure is a `HashMap`. Probably import assertions should be stored as an `IndexMap` so it's stable.

nayeemrmn: Done. And since errors are unstructured strings, this doesn't stop us from still including assertion errors in `Import::errors`.

dsherret (approved):
LGTM. Thanks @nayeemrmn!

PR #271, opened 2023-04-23
fix: check all import ranges in Dependency::includes()

nayeemrmn:
This is preventing the LSP hover text on repeated imports from including certain dependency information: https://github.com/denoland/deno/blob/5a524a9a5a7fac0d16b2cbe2df1142dc419df7fc/cli/lsp/documents.rs#L603.


dsherret (approved):
LGTM. I will kick off a patch release after merging

PR #295, opened 2023-09-20
feat: resolve node builtin module without node: scheme

kt3k:
part of https://github.com/denoland/deno/issues/20566

This PR updates the resolution of bare specifiers. If the module specifier is node builtin module as a bare specifier, resolve it as `node:<module_name>`. Also it prints the warning message like `Warning: Resolving "path" as "node:path". If you want to use a built-in Node module, add a "node:" prefix.`

### DIFF
--- a/lib/lib.rs
+++ b/lib/lib.rs
@@ -156,7 +156,7 @@ impl Resolver for JsResolver {
       ModuleSpecifier::parse(&value)
         .map_err(|err| ResolveError::Specifier(SpecifierError::InvalidUrl(err)))
     } else {
-      resolve_import(specifier, referrer, None).map_err(|err| err.into())
+      resolve_import(specifier, referrer).map_err(|err| err.into())
     }
   }
 
--- a/lib/lib.rs
+++ b/lib/lib.rs
@@ -295,6 +295,7 @@ pub fn js_parse_module(
     content.into(),
     maybe_resolver.as_ref().map(|r| r as &dyn Resolver),
     None,
+    None,
   ) {
     Ok(module) => {
       let serializer =
--- a/lib/lib.rs
+++ b/lib/lib.rs
@@ -155,7 +155,7 @@ impl Resolver for JsResolver {
       };
       Ok(ModuleSpecifier::parse(&value)?)
     } else {
-      resolve_import(specifier, referrer).map_err(|err| err.into())
+      resolve_import(specifier, referrer, None).map_err(|err| err.into())
     }
   }
 
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -824,6 +824,7 @@ impl GraphImport {
     referrer: &ModuleSpecifier,
     imports: Vec<String>,
     maybe_resolver: Option<&dyn Resolver>,
+    maybe_npm_resolver: Option<&dyn NpmResolver>,
   ) -> Self {
     let dependencies = imports
       .into_iter()
@@ -833,7 +834,8 @@ impl GraphImport {
           start: Position::zeroed(),
           end: Position::zeroed(),
         };
-        let maybe_type = resolve(&import, referrer_range, maybe_resolver);
+        let maybe_type =
+          resolve(&import, referrer_range, maybe_resolver, maybe_npm_resolver);
         (
           import,
           Dependency {
@@ -1552,12 +1554,17 @@ fn resolve(
   specifier_text: &str,
   referrer_range: Range,
   maybe_resolver: Option<&dyn Resolver>,
+  maybe_npm_resolver: Option<&dyn NpmResolver>,
 ) -> Resolution {
   let response = if let Some(resolver) = maybe_resolver {
     resolver.resolve(specifier_text, &referrer_range.specifier)
   } else {
-    resolve_import(specifier_text, &referrer_range.specifier, None)
-      .map_err(|err| err.into())
+    resolve_import(
+      specifier_text,
+      &referrer_range.specifier,
+      maybe_npm_resolver,
+    )
+    .map_err(|err| err.into())
   };
   Resolution::from_resolve_result(response, specifier_text, referrer_range)
 }
@@ -1622,6 +1629,7 @@ pub(crate) fn parse_module(
   module_analyzer: &dyn ModuleAnalyzer,
   is_root: bool,
   is_dynamic_branch: bool,
+  maybe_npm_resolver: Option<&dyn NpmResolver>,
 ) -> Result<Module, ModuleGraphError> {
   let media_type =
     MediaType::from_specifier_and_headers(specifier, maybe_headers);
@@ -1688,6 +1696,7 @@ pub(crate) fn parse_module(
             module_info,
             content,
             maybe_resolver,
+            maybe_npm_resolver,
           )))
         }
         Err(diagnostic) => Err(ModuleGraphError::ModuleError(
@@ -1710,6 +1719,7 @@ pub(crate) fn parse_module(
             module_info,
             content,
             maybe_resolver,
+            maybe_npm_resolver,
           )))
         }
         Err(diagnostic) => Err(ModuleGraphError::ModuleError(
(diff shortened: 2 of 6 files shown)


kt3k:
Tested with the below change in Deno CLI repo:

```diff
index 36dccab10..6bbdc96c0 100644
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -49,16 +49,16 @@ deno_ast = { workspace = true, features = ["bundler", "cjs", "codegen", "dep_gra
 deno_cache_dir = "=0.6.0"
 deno_config = "=0.3.1"
 deno_core = { workspace = true, features = ["include_js_files_for_snapshotting"] }
-deno_doc = "=0.67.0"
-deno_emit = "=0.28.0"
-deno_graph = "=0.55.0"
+deno_doc = { path = "../../deno_doc" }
+deno_emit = { path = "../../deno_emit/rs-lib" }
+deno_graph = { path = "../../deno_graph" }
 deno_lint = { version = "=0.51.0", features = ["docs"] }
 deno_lockfile.workspace = true
 deno_npm.workspace = true
 deno_runtime = { workspace = true, features = ["dont_create_runtime_snapshot", "exclude_runtime_main_js", "include_js_files_for_snapshotting"] }
 deno_semver.workspace = true
 deno_task_shell = "=0.13.2"
-eszip = "=0.53.0"
+eszip = { path = "../../eszip" }
 napi_sym.workspace = true
 
 async-trait.workspace = true
```

It works like the below:
```shellsession
$ cat a.js                       
import fs from "fs";
import path from "path";
console.log(fs.write);
$ ./target/debug/deno run a.js 
Warning: Resolving "fs" as "node:fs". If you want to use a built-in Node module, add a "node:" prefix.
Warning: Resolving "path" as "node:path". If you want to use a built-in Node module, add a "node:" prefix.
[Function: write]
```

src/module_specifier.rs
```diff
@@ -57,10 +116,19 @@ pub fn resolve_import(
...
+    //    If the result is not failure, return the result.
     Ok(url) => url,
 
-    // 2. If specifier does not start with the character U+002F SOLIDUS (/),
+    // 2. If specifier is a node specifier without node: scheme, return
+    //    the specifier with node: scheme added.
+    Err(ParseError::RelativeUrlWithoutBase)
+      if is_builtin_node_module(specifier) =>
+    {
+      eprintln!("Warning: Resolving \"{specifier}\" as \"node:{specifier}\". If you want to use a built-in Node module, add a \"node:\" prefix.");
```
dsherret, line 127: We should be able to disable this warning via the `--quiet` flag in the CLI. Maybe we could have an additional method in `NpmResolver` for allowing the implementor to log these errors. Alternatively, we could just use the `log` crate in deno_graph. There are some other `eprintln`s that we could replace with the log crate while doing that.

dsherret (changes requested):
@kt3k looks mostly good, but we need to make some changes.

src/module_specifier.rs
```diff
@@ -49,6 +49,65 @@ fn specifier_from_path(_path: PathBuf) -> ModuleSpecifier {
   ModuleSpecifier::parse(EMPTY_SPECIFIER).unwrap()
 }
 
+static SUPPORTED_BUILTIN_NODE_MODULES: &[&str] = &[
```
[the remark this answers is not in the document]
kt3k: `resolve_builtin_node_module` accepts `ModuleSpecifier` (`Url`) as parameter, but in this change we'd like to check if the given `str` is a bare specifier specifying builtin node module. I don't think we can use `resolve_builtin_node_module` as is for that purpose.

Does it make sense to add new method (like `is_builtin_node_module(module_name: &str) -> bool`) to `NpmResolver`? Or should we create a new trait for that specific purpose?

kt3k:
Now this change uses `is_builtin_node_module_name` and `on_resolve_bare_builtin_node_module` from `NpmResolver` trait for checking builtin node module and printing warnings.

I skipped supporting the same thing with wasm/js version as npm resolver doesn't seem configured for them.

src/module_specifier.rs
```diff
@@ -54,13 +55,27 @@ fn specifier_from_path(_path: PathBuf) -> ModuleSpecifier {
...
-    //    If the result is not failure, return he result.
+    //    If the result is not failure, return the result.
     Ok(url) => url,
 
-    // 2. If specifier does not start with the character U+002F SOLIDUS (/),
+    // 2. If specifier is a node specifier without node: scheme, return
+    //    the specifier with node: scheme added.
+    Err(ParseError::RelativeUrlWithoutBase)
+      if maybe_npm_resolver.map_or(false, |npm_resolver| {
+        npm_resolver.is_builtin_node_module_name(specifier)
```
dsherret, line 69: Perhaps instead we could use:

```rs
// probably doesn't compile, but this is the basic idea
ModuleSpecifier::parse(format!("node:{}", specifier)).ok().and_then(|specifier| npm_resolver.resolve_builtin_node_module(&specifier).ok().flatten()).is_some()
```

Then we wouldn't need a separate `is_builtin_node_module_name` function

src/module_specifier.rs
```diff
@@ -54,13 +55,27 @@ fn specifier_from_path(_path: PathBuf) -> ModuleSpecifier {
...
+      }) =>
+    {
```
dsherret, line 71: I think we should move this check to a higher level in:

https://github.com/denoland/deno_graph/blob/aa9236c40a2b8e2514c1f08e399b2c0ec70f81fd/src/graph.rs#L1551-L1562

https://github.com/denoland/deno_graph/pull/296 will make this easier.

That way this behaviour isn't dependent on an implentor of the `Resolver` calling this `resolve_import` function and needing access to the npm resolver. Basically, if a user provides an `NpmResolver` to deno_graph then they'll get this behaviour.

lib/lib.rs
dsherret, line 1: I know we discussed making this conditional in the meeting, but maybe making it "just work" is ok to do in a patch release of Deno. It would be less work and make stuff work sooner. cc @bartlomieju

dsherret (commented):
Very close now. Looking good!

src/module_specifier.rs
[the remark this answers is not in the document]
dsherret: Let's discuss in https://github.com/denoland/deno_graph/pull/295#discussion_r1337472686

lib/lib.rs
bartlomieju: That's okay for me 👍

src/module_specifier.rs
kt3k: Thanks! Removed `is_builtin_node_module_name`

src/module_specifier.rs
kt3k: Moved the check of bare builtin node specifier. Now it is checked after `resolver.resolve` is applied.

dsherret (approved):
Nice! LGTM!

src/graph.rs
```diff
@@ -1550,13 +1552,33 @@ fn resolve(
...
+    if let Some(npm_resolver) = maybe_npm_resolver {
+      let maybe_specifier =
+        ModuleSpecifier::parse(&format!("node:{}", specifier_text)).ok();
+      let maybe_mod_name = maybe_specifier.as_ref().and_then(|s| {
+        npm_resolver.resolve_builtin_node_module(s).ok().flatten()
+      });
+      if maybe_mod_name.is_some() {
+        npm_resolver.on_resolve_bare_builtin_node_module(specifier_text);
+        return Resolution::from_resolve_result(
+          Ok(maybe_specifier.unwrap()),
```
dsherret, line 1575: FWIW, the `.unwrap()` could be removed by using `.filter` on the result of `ModuleSpecifier::parse(&format!("node:{}", specifier_text)).ok()` when `npm_resolver.resolve_builtin_node_module(s).ok().flatten().is_some()`

kt3k: > FWIW, the .unwrap() could be removed by using .filter on the result of ModuleSpecifier::parse(&format!("node:{}", specifier_text)).ok()

I didn't get what this exactly means.., but removed the usage of `.unwrap()` in a different way. https://github.com/denoland/deno_graph/pull/295/commits/4ebda530c0c60584707c9a8aee3f74d14f0457f4

kt3k:
@dsherret Thank you for your review! I'll land this and release it as v0.56.0 tomorrow my time (and will continue working on dependent crates, such as `eszip`, `deno_emit`, `deno_doc`, etc).

kt3k:
https://github.com/denoland/eszip/pull/149
https://github.com/denoland/deno_doc/pull/352
https://github.com/denoland/deno_emit/pull/140

PR #297, opened 2023-09-29
fix: pass range to on_resolve_bare_builtin_node_module

kt3k:
This PR passes `Range` to `on_resolve_bare_builtin_node_module`. This is necessary to implement https://github.com/denoland/deno/pull/20728

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1093,11 +1093,17 @@ console.log(a);
       }
     }
 
-    fn on_resolve_bare_builtin_node_module(&self, module_name: &str) {
-      eprintln!(
-        "Warning: Resolving bare specifier \"{}\" to \"node:{}\".",
-        module_name, module_name
-      );
+    fn on_resolve_bare_builtin_node_module(
+      &self,
+      module_name: &str,
+      range: &Range,
+    ) {
+      let Range {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1093,11 +1093,17 @@ console.log(a);
...
+      &self,
+      module_name: &str,
+      range: &Range,
+    ) {
+      let Range {
+        specifier, start, ..
+      } = range;
+      let line = start.line + 1;
+      let column = start.character;
+      eprintln!("Warning: Resolving \"{module_name}\" as \"node:{module_name}\" at {specifier}:{line}:{column}. If you want to use a built-in Node module, add a \"node:\" prefix.");
```
kt3k, line 1106: This prints:
```
Warning: Resolving "path" as "node:path" at file:///a/test.ts:1:7. If you want to use a built-in Node module, add a "node:" prefix.
```

dsherret (approved):
Great! We don't use this function in eszip, deno_doc, or deno_emit, right? Let's just do a patch release because for sure nobody is using this functionality yet and that will allow us to not have to bump the deno_graph version in the other crates (so we can merge this then immediately update the cli)

PR #298, opened 2023-10-02
fix: check bare node specifier for ImportMapError::UnmappedBareSpecifier

kt3k:
This PR adds bare node specifier check when the `ResolveError` from `resolver.resolve` is `Other(ImportMapError::UnmappedBareSpecifier(_, _))`.

By this change, we can keep more detailed error message from `ImportMapError`, for example, like `Relative import path "unmapped" not prefixed with / or ./ or ../ and not in import map from "file://[WILDCARD]/092_import_map_unmapped_bare_specifier.ts"`

related https://github.com/denoland/deno/pull/20728

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,6 +171,7 @@ dependencies = [
  "deno_ast",
  "deno_semver",
  "futures",
+ "import_map",
  "indexmap 2.0.0",
  "monch",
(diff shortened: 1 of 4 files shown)


dsherret (approved):
LGTM!

PR #299, opened 2023-10-04
fix: add is_bare_node_specifier_enabled to NpmResolver trait

kt3k:
This PR adds `is_bare_node_specifier_enabled` to `NpmResolver` to enable feature flagging of bare node specifier resolution.

related https://github.com/denoland/deno/pull/20728


dsherret (approved):
LGTM! We can just do a patch release again to avoid updating the other crates.

PR #319, opened 2023-11-04
Rename types.d.ts to type.ts

yacinehmito:
Declaration files should only be used to represent types that result from the transpilation of TypeScript to JavaScript. In a TypeScript context, it can cause issues that, from my understanding, relate to ambient types.


dsherret (approved):
LGTM. Thanks!

PR #359, opened 2024-01-19
refactor: improve handling of FastCheck diagnostics

lucacasonato:


src/fast_check/mod.rs
```diff
@@ -26,58 +27,61 @@ fn format_diagnostics(diagnostics: &[FastCheckDiagnostic]) -> String {
...
-  MissingExplicitReturnType { range: Range },
-  #[error("Global augmentations such as ambient modules are not supported.")]
-  UnsupportedAmbientModule { range: Range },
-  #[error("The reference '{name}' from '{referrer}' was too complex. Extract out the shared type to a type alias.")]
+  #[error("missing explicit type in the public API")]
+  MissingExplicitType { range: DiagnosticRange },
+  #[error("missing explicit return type in the public API")]
+  MissingExplicitReturnType { range: DiagnosticRange },
+  #[error(
+    "found an ambient module, which is a global augmentation, which are not unsupported "
```
dsherret, line 41: Extra space on end?

src/fast_check/mod.rs
```diff
@@ -26,58 +27,61 @@ fn format_diagnostics(diagnostics: &[FastCheckDiagnostic]) -> String {
 
 #[derive(Debug, Clone, thiserror::Error)]
 pub enum FastCheckDiagnostic {
-  #[error("Could not resolve '{name}' referenced from '{referrer}'. This may indicate a bug in Deno. Please open an issue to help us improve if so.")]
+  #[error("could not resolve '{name}' referenced from '{referrer}'")]
```
dsherret, line 30: I'm not really fan of lowercase error messages and most of our error messages don't do this (typescript diagnostics and v8 errors also don't do it). I understand `thiserror` recommends it and the rust std library does it, but I don't get why.

src/fast_check/mod.rs
```diff
@@ -119,23 +147,25 @@ impl FastCheckDiagnostic {
...
       UnsupportedTsNamespaceExport { range } => Some(range),
       UnsupportedUsing { range } => Some(range),
-      UnsupportedJavaScriptEntrypoint { range } => Some(range),
-      UnsupportedNestedJavaScript { range } => Some(range),
-      Emit(_) => None,
-      Multiple(_) => None,
+      UnsupportedJavaScriptEntrypoint { .. } => None,
+      UnsupportedNestedJavaScript { .. } => None,
+      Emit { .. } => None,
+      Multiple(_) => unreachable!("Multiple should be flattened"),
```
dsherret, line 153: Maybe debug assertion only, then return `None` in release? It's a bug, but I don't think worth taking down the process for.

src/fast_check/range_finder.rs
```diff
@@ -370,11 +369,7 @@ impl<'a> PublicRangeFinder<'a> {
           );
         } else {
           errors.push(FastCheckDiagnostic::UnsupportedJavaScriptEntrypoint {
-            range: Range {
-              start: crate::Position::zeroed(),
-              end: crate::Position::zeroed(),
-              specifier: specifier.clone(),
-            },
+            specifier: specifier.clone(),
```
dsherret, line 372: 👍

src/fast_check/mod.rs
```diff
@@ -26,58 +27,61 @@ fn format_diagnostics(diagnostics: &[FastCheckDiagnostic]) -> String {
...
     name: String,
     referrer: String,
   },
-  #[error("Missing explicit type in the public API.")]
-  MissingExplicitType { range: Range },
-  #[error("Missing explicit return type in the public API.")]
-  MissingExplicitReturnType { range: Range },
-  #[error("Global augmentations such as ambient modules are not supported.")]
-  UnsupportedAmbientModule { range: Range },
-  #[error("The reference '{name}' from '{referrer}' was too complex. Extract out the shared type to a type alias.")]
```
dsherret, line 41: Where are the recommendations for this stored? We should probably keep that in deno_graph rather than elsewhere so that someone can more easily get a diagnostic with the recommendation even if they're not using the CLI.

src/fast_check/mod.rs
lucacasonato: these error messages are prefixed by `error[zap-not-found-reference]:` in output. I think not using caps helps the text flow better (looks like one sentence then).

src/fast_check/mod.rs
lucacasonato: Ok - then I'll move hints, descriptions, etc all into here.

src/fast_check/mod.rs
dsherret: Ok, that's cool if it's going away.

crowlKats (approved):
RSLGTM

PR #366, opened 2024-01-24
feat: support leaving arrows and fns in subset type checking

### LINKED ISSUE #365: Support inferring function signature types in subset type checking
Currently each of the following exports errors with a "missing explicit type annotation" error. Instead subset type checking should support these types, performing the same function rewriting as we do for function declarations and code.

```ts
export const test1 = function (): string { return "string" };
export const test2 = (): string => "string";
```

As a bonus, for arrow functions with an expression body, we could support the same simple inference that we do for RHS of var decls. For example:

```ts
export const test3 = () => "test" as const;
```

Possible output for these:

```ts
export const test1 = function (): string { return {} as any };
export const test2 = (): string => ({} as any);
export const test3 = () => "test" as const;
```

or, if we do this via `maybe_infer_type_from_expr`

```ts
export const test1: () => string = {} as any;
export const test2: () => string = {} as any;
// test3 errors with "missing explicit return type"
```

lucacasonato:
Fixes #365

### DIFF
--- a/src/fast_check/transform.rs
+++ b/src/fast_check/transform.rs
@@ -11,6 +11,7 @@ use deno_ast::swc::ast::Accessibility;
 use deno_ast::swc::ast::ArrayLit;
 use deno_ast::swc::ast::ArrowExpr;
 use deno_ast::swc::ast::BindingIdent;
+use deno_ast::swc::ast::BlockStmt;
 use deno_ast::swc::ast::BlockStmtOrExpr;
 use deno_ast::swc::ast::CallExpr;
 use deno_ast::swc::ast::Callee;
@@ -853,25 +854,9 @@ impl<'a> FastCheckTransformer<'a> {
       if let Some(body) = &mut n.body {
         let return_stmts = get_return_stmts_with_arg_from_function_body(body);
         if return_stmts.is_empty() {
-          let void_type =
-            Box::new(ts_keyword_type(TsKeywordTypeKind::TsVoidKeyword));
           n.return_type = Some(Box::new(TsTypeAnn {
             span: DUMMY_SP,
-            type_ann: if n.is_async {
-              Box::new(TsType::TsTypeRef(TsTypeRef {
-                span: DUMMY_SP,
-                type_name: TsEntityName::Ident(Ident::new(
-                  "Promise".into(),
-                  DUMMY_SP,
-                )),
-                type_params: Some(Box::new(TsTypeParamInstantiation {
-                  span: DUMMY_SP,
-                  params: vec![void_type],
-                })),
-              }))
-            } else {
-              void_type
-            },
+            type_ann: void_or_promise_void(n.is_async),
           }));
         } else {
           self.mark_diagnostic(
@@ -886,25 +871,16 @@ impl<'a> FastCheckTransformer<'a> {
     if let Some(body) = &mut n.body {
       body.stmts.clear();
 
-      // push a return stmt to suppress type errors
-      let has_void_return_type = n
+      // push a return stmt to suppress type errors, if the function is not a
+      // void function
+      let return_expr = n
         .return_type
         .as_ref()
-        .map(|t| is_void_type(&t.type_ann))
-        .unwrap_or(true);
-      if !has_void_return_type {
-        let has_never_return_type = n
-          .return_type
-          .as_ref()
-          .map(|t| is_never_type(&t.type_ann))
-          .unwrap_or(false);
+        .and_then(|t| replacement_return_value(&t.type_ann));
+      if let Some(return_expr) = return_expr {
         body.stmts.push(Stmt::Return(ReturnStmt {
           span: DUMMY_SP,
-          arg: Some(if has_never_return_type {
-            obj_as_never_expr()
-          } else {
-            obj_as_any_expr()
-          }),
+          arg: Some(return_expr),
         }));
       }
     }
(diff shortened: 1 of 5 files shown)


lucacasonato (commented):
The new `maybe_transform_expr_if_leavable` function is mostly a copy of `is_expr_leavable` from `swc_helpers.rs`. I was not sure how to best de-dupe this. David, do you have an idea?

src/fast_check/transform.rs
```diff
@@ -903,6 +921,101 @@ impl<'a> FastCheckTransformer<'a> {
...
+    &mut self,
+    n: &mut ArrowExpr,
+    parent_id_range: Option<SourceRange>,
+  ) -> Result<(), Box<FastCheckDiagnostic>> {
+    let range = parent_id_range.unwrap_or_else(|| n.range());
+
+    let mut clear_body = true;
+
+    if n.return_type.is_none() {
+      if n.is_generator {
```
dsherret, line 934: Do we have a test for this?

src/fast_check/transform.rs
```diff
@@ -903,6 +921,101 @@ impl<'a> FastCheckTransformer<'a> {
...
+                  )),
+                  type_params: Some(Box::new(TsTypeParamInstantiation {
+                    span: DUMMY_SP,
+                    params: vec![void_type],
+                  })),
+                }))
+              } else {
+                void_type
+              },
+            }));
```
dsherret, line 965: This code seems the same as above? Maybe we could add some helper functions at the bottom of this file for more easily creating these `TsTypeAnn`s.

src/fast_check/transform.rs
```diff
@@ -1194,6 +1430,7 @@ impl<'a> FastCheckTransformer<'a> {
           None
         }
       }
+      Expr::Paren(n) => self.maybe_infer_type_from_expr(&n.expr),
```
dsherret, line 1433: Is there a test for this?

tests/specs/graph/JsrSpecifiers_FastCheck_Vars.txt
```diff
@@ -35,6 +35,11 @@ export const asIs2 = {
   g: 1++,
   h: [...[1, 2], ...["test"]],
 };
+export const func1 = function (): string { return "test" };
+export const func2 = (): string => "test";
+export const func3 = () => "test";
+export const func4 = () => "test" as const;
```
dsherret, line 41: We should add a test like the following in order to ensure the dep analysis is working for functions and arrows in variables too:

```ts
export const func5 = (arg: Public1) => "test" as const;
export const func6 = function(): Public2 { return new Public2(); };

interface Public1 {}
class Public2 {}
```

tests/specs/graph/JsrSpecifiers_FastCheck_ClassPropertiesLeavable.txt
```diff
@@ -0,0 +1,120 @@
...
+# https://jsr.io/@scope/a/1.0.0_meta.json
+{ "exports": { ".": "./mod.ts" } }
+
+# https://jsr.io/@scope/a/1.0.0/mod.ts
+import type { Foo, Bar } from "./b.ts";
+
+export class Foo {
+  bar = (a: string) => "test";
+  foo = () => "test" as Foo;
+  bar = (b: Bar) => 1;
```
dsherret, line 13: We should add a similar test for function expressions. Something like:

```ts
  baz = function(): PublicByBaz { return new PublicByBaz(); }
}

class PublicByBaz {}
```

src/fast_check/transform.rs
```diff
@@ -1154,6 +1277,119 @@ impl<'a> FastCheckTransformer<'a> {
     DiagnosticRange::new(self.specifier.clone(), range)
   }
 
+  fn maybe_transform_expr_if_leavable(
```
dsherret, line 1280: Hmmm yeah tricky. I believe that the mutation is only necessary for if the expr is an `Arrow` or `Fn`, correct? It seems tempting to only support those as leavable at the top level so that we could re-use is_expr_leavable otherwise, but it also seems nice to do this recursrively as well. Perhaps we should just add a bit warning comment to both functions that we should keep them in sync?

dsherret (approved):
Amazing! Thanks!

src/fast_check/transform.rs
lucacasonato: I just realized generator arrow functions are not actually a thing. What is this even lol - why does SWC have this prop in the AST hahaha

src/fast_check/transform.rs
lucacasonato: Yes - https://github.com/denoland/deno_graph/blob/e22155973b1d687fcdcb984154919e59873c1bb6/tests/specs/graph/JsrSpecifiers_FastCheck_Vars.txt#L56

src/fast_check/transform.rs
dsherret: Oh, I didn't realize this was for an arrow function. Often invalid states are stored in the ast for better diagnostics.

tests/specs/graph/JsrSpecifiers_FastCheck_ClassPropertiesLeavable.txt
lucacasonato: Done

lucacasonato:
While implementing your requests I realized my return type inference for arrow functions with expr bodies does not take into account `async` functions yet - will fix this asap

src/fast_check/transform.rs
lucacasonato: de-duped

src/fast_check/transform.rs
lucacasonato: i'll add a comment

Issue #364, opened 2024-01-23
Standard way of resolving import maps


deer:
Is there no standard way of resolving import maps?

I have a simple project:
```
tree
.
├── a.ts
├── b.ts
└── deno.jsonc
```

```ts
//a.ts
import { foo } from "hiddenBehindMap";

console.log(foo);
```

```ts
//b.ts
export const foo = "foo";
```

```jsonc
//deno.jsonc
{
  "imports": {
    "hiddenBehindMap": "./b.ts"
  }
}
```

And then I have a slightly less simple test file:
```ts
import { createGraph } from "https://deno.land/x/deno_graph@0.63.5/mod.ts";
import { resolve, toFileUrl } from "https://deno.land/std@0.212.0/path/mod.ts";

const path = resolve("./a.ts");
const fileUrlString = toFileUrl(path).toString();
let graph = await createGraph(fileUrlString);
console.log(graph);

const imports = (await import("./deno.json", { with: { type: "json" } }))
  .default.imports as Record<string, string>;
graph = await createGraph(fileUrlString, {
  resolve: createCustomResolver(imports),
});
console.log(graph);

function createCustomResolver(imports: Record<string, string>) {
  return (specifier: string, referrer: string) => {
    for (const key of Object.keys(imports)) {
      if (specifier.startsWith(key)) {
        specifier = specifier.replace(key, imports[key]);
        break;
      }
    }
    return new URL(specifier, referrer).toString();
  };
}
```
When I run the above I get the following output:
```
{
  roots: [ "file:///Users/reed/code/denoland/deno_graph/test_issue/a.ts" ],
  modules: [
    {
      kind: "esm",
      dependencies: [ { specifier: "hiddenBehindMap", code: [Object] } ],
      size: 90,
      mediaType: "TypeScript",
      specifier: "file:///Users/reed/code/denoland/deno_graph/test_issue/a.ts"
    }
  ],
  redirects: {}
}
{
  roots: [ "file:///Users/reed/code/denoland/deno_graph/test_issue/a.ts" ],
  modules: [
    {
      kind: "esm",
      dependencies: [ { specifier: "hiddenBehindMap", code: [Object] } ],
      size: 90,
      mediaType: "TypeScript",
      specifier: "file:///Users/reed/code/denoland/deno_graph/test_issue/a.ts"
    },
    {
      kind: "esm",
      size: 25,
      mediaType: "TypeScript",
[...]

PR #369, opened 2024-01-31
chore: add documentation about using import_map with deno_graph

deer:
closes https://github.com/denoland/deno_graph/issues/364

I found the usage of `deno_graph` to be unexpected, and it was a long journey to get where I wanted. Having an example like this (with a focus on using `import_map`) would have saved me a ton of time. I've thoroughly prepared this example with the hope that it helps someone else in the future.

This can be more properly viewed here:
https://github.com/deer/deno_graph/tree/docs_for_import_map/js#exploring-the-entire-dependency-graph


js/README.md
```diff
@@ -43,3 +43,183 @@ There are several options that can be passed the function in the optional
...
+library provides a `resolve` function which we can pass to `createGraph`.
+
+```ts
+import { createGraph } from "https://deno.land/x/deno_graph@0.63.6/mod.ts";
+import { resolve, toFileUrl } from "https://deno.land/std@0.212.0/path/mod.ts";
+import { parseFromJson } from "https://deno.land/x/import_map@v0.18.3/mod.ts";
+
+const path = resolve("./someFiles/c.ts");
+const importMap =
+  (await import("./deno.json", { with: { type: "json" } })).default;
```
dsherret, line 173: This won't work for jsonc deno.json files. Maybe this should use `Deno.readTextFileSync(new URL("./deno.json", import.meta.url))` instead?

dsherret (commented):
Thanks for this PR! It's a little long. Maybe we should just show the code example of how to use it with /x/import_map?

deer:
Thanks for looking. It's much shorter now.

js/README.md
```diff
@@ -43,3 +43,27 @@ There are several options that can be passed the function in the optional
...
+);
+const resolvedImportMap = await parseFromJson(
+  toFileUrl(resolve("deno.json")),
+  importMap,
+);
+const graph = await createGraph(toFileUrl(path).href, {
+  resolve: resolvedImportMap.resolve.bind(resolvedImportMap),
+});
+console.log(graph);
+```
```
lino-levan, line 69: Could you add comments here? The code is relatively clear but I wish it was more clearly explained.

lino-levan (changes requested):

js/README.md
deer: I added some comments. Let me know what you think.

dsherret (approved):
Thanks! I updated the docs for the new API on JSR (the import map crate has yet to migrate to JSR, so leaving that url there for now)

PR #378, opened 2024-02-12
fix: keep @ts- comments

### LINKED ISSUE #377: Do not remove `// @ts-ignore` comments in zap
As title.

lino-levan:
Closes #377.

### DIFF
--- a/src/fast_check/transform.rs
+++ b/src/fast_check/transform.rs
@@ -110,11 +110,8 @@ impl CommentsMut {
       comments.retain(|_key, value| {
         value.retain(|c| {
           match c.kind {
-            // only keep js docs, @ts-*, and @deno-types comments
-            CommentKind::Line => {
-              let text = c.text.trim_start();
-              text.starts_with("@deno-types") || text.starts_with("@ts-")
-            }
+            // only keep js docs and @ts-* comments
+            CommentKind::Line => c.text.trim_start().starts_with("@ts-"),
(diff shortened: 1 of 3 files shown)


src/fast_check/transform.rs
```diff
@@ -110,8 +110,10 @@ impl CommentsMut {
       comments.retain(|_key, value| {
         value.retain(|c| {
           match c.kind {
-            // only keep js docs and @deno-types comments
-            CommentKind::Line => c.text.starts_with("@deno-types"),
+            // only keep js docs, @ts-*, and @deno-types comments
+            CommentKind::Line => {
+              c.text.starts_with("@deno-types") || c.text.starts_with("@ts-")
```
dsherret, line 115: Oh, this actually might need to use `.trim_start()` first for both of these I think.

dsherret (commented):
Thanks! Would you be able to add a test in:

https://github.com/denoland/deno_graph/tree/main/tests/specs/graph

1. Copy and paste one of the other files like _FastCheck_ to a new file.
2. Modify the source to add an example that uses these comments.
3. Run with `UPDATE=1 cargo test --all-features` to update the spec file output and see if it worked.

src/fast_check/transform.rs
lino-levan: We should definitely have a test for that.

dsherret (approved):
LGTM. Thanks!

PR #379, opened 2024-02-13
feat: Add fast-check based TypeScript type definition generation

marvinhagemeister:
This PR adds the ability to generate TypeScript definition files (`*.d.ts`) when fast check is enabled.

### DIFF
--- a/src/fast_check/transform_dts.rs
+++ b/src/fast_check/transform_dts.rs
@@ -270,49 +270,19 @@ impl<'a> FastCheckDtsTransformer<'a> {
           prev_is_overload = false;
           if let Stmt::Decl(decl) = stmt {
             match decl {
-              Decl::TsEnum(ts_enum) => {
-                if let Some(decl) =
-                  self.decl_to_type_decl(Decl::TsEnum(ts_enum.clone()))
-                {
+              Decl::TsEnum(_) | Decl::Class(_) | Decl::Fn(_) | Decl::Var(_) => {
+                if let Some(decl) = self.decl_to_type_decl(decl.clone()) {
                   new_items.push(ModuleItem::Stmt(Stmt::Decl(decl)));
                 } else {
-                  self.mark_diagnostic_unable_to_infer(ts_enum.range())
+                  self.mark_diagnostic_unable_to_infer(decl.range())
                 }
               }
+
               Decl::TsInterface(_)
               | Decl::TsTypeAlias(_)
               | Decl::Using(_)
               | Decl::TsModule(_) => {
                 new_items.push(ModuleItem::Stmt(Stmt::Decl(decl)));
-              }
-              // Since fast check requires explicit type annotations we
-              // can drop other statements not part of an export statement.
-              Decl::Class(class_decl) => {
-                if let Some(decl) =
-                  self.decl_to_type_decl(Decl::Class(class_decl.clone()))
-                {
-                  new_items.push(ModuleItem::Stmt(Stmt::Decl(decl)));
-                } else {
-                  self.mark_diagnostic_unable_to_infer(class_decl.range())
-                }
-              }
-              Decl::Fn(fn_decl) => {
-                if let Some(decl) =
-                  self.decl_to_type_decl(Decl::Fn(fn_decl.clone()))
-                {
-                  new_items.push(ModuleItem::Stmt(Stmt::Decl(decl)));
-                } else {
-                  self.mark_diagnostic_unable_to_infer(fn_decl.range())
-                }
-              }
-              Decl::Var(var_decl) => {
-                if let Some(decl) =
-                  self.decl_to_type_decl(Decl::Var(var_decl.clone()))
-                {
-                  new_items.push(ModuleItem::Stmt(Stmt::Decl(decl)));
-                } else {
-                  self.mark_diagnostic_unable_to_infer(var_decl.range())
-                }
               }
             }
           }
--- a/src/fast_check/transform_dts.rs
+++ b/src/fast_check/transform_dts.rs
@@ -147,7 +147,8 @@ impl FastCheckDtsTransformer {
         })))
       }
       Expr::Object(obj) => {
-        let mut members: Vec<TsTypeElement> = Vec::with_capacity(obj.len());
+        let mut members: Vec<TsTypeElement> =
+          Vec::with_capacity(obj.props.len());
 
         for item in obj.props {
           match item {
--- a/src/fast_check/transform_dts.rs
+++ b/src/fast_check/transform_dts.rs
@@ -7,7 +7,7 @@ use deno_ast::swc::{
     TsTupleElement, TsTupleType, TsType, TsTypeElement, TsTypeLit, VarDecl,
     VarDeclKind, VarDeclarator,
   },
-  common::{Spanned, DUMMY_SP},
+  common::DUMMY_SP,
 };
 
 use crate::FastCheckDiagnostic;
@@ -35,7 +35,7 @@ impl FastCheckDtsTransformer {
     &mut self,
     mut module: Module,
   ) -> Result<Module, Vec<FastCheckDiagnostic>> {
-    let mut body = module.body;
+    let body = module.body;
 
     let mut new_items: Vec<ModuleItem> = vec![];
 
@@ -59,7 +59,7 @@ impl FastCheckDtsTransformer {
             let value = match export_decl.decl {
               DefaultDecl::Class(mut class_expr) => {
                 class_expr.class.body =
-                  self.class_body_to_type(&class_expr.class.body);
+                  self.class_body_to_type(class_expr.class.body);
                 ExportDefaultDecl {
                   span: export_decl.span,
                   decl: DefaultDecl::Class(class_expr),
@@ -83,7 +83,7 @@ impl FastCheckDtsTransformer {
             let name = self.gen_unique_name();
             let name_ident = Ident::new(name.into(), DUMMY_SP);
             let type_ann = self
-              .expr_to_ts_type(*export_default_expr.expr.clone(), false)
+              .expr_to_ts_type(*export_default_expr.expr, false)
               .map(type_ann)
               .or(Some(any_type_ann()));
 
@@ -114,10 +114,11 @@ impl FastCheckDtsTransformer {
           // Keep all these
           ModuleDecl::TsImportEquals(_)
           | ModuleDecl::TsNamespaceExport(_)
-          | ModuleDecl::TsExportAssignment(_) => {
-            new_items.push(ModuleItem::ModuleDecl(module_decl.clone()));
+          | ModuleDecl::TsExportAssignment(_)
+          | ModuleDecl::ExportNamed(_)
+          | ModuleDecl::ExportAll(_) => {
+            new_items.push(ModuleItem::ModuleDecl(module_decl));
           }
-          _ => {}
         },
         ModuleItem::Stmt(stmt) => {
           if let Stmt::Decl(decl) = stmt {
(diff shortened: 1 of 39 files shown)


src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,730 @@
...
+    format!("_dts_{}", self.id_counter)
+  }
+
+  pub fn transform(
+    &mut self,
+    mut module: Module,
+  ) -> Result<Module, Vec<FastCheckDiagnostic>> {
+    let mut body = module.body;
+
+    let mut new_items: Vec<ModuleItem> = vec![];
```
dsherret, line 39: We can pre-allocate this:

```suggestion
    let mut new_items: Vec<ModuleItem> = Vec::with_capacity(body.len());
```

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,730 @@
...
+
+    for item in &mut body {
```
dsherret, line 41: Maybe we can consume the body instead and to reduce cloning? `for item in body`

It will require being a bit more verbose below, but it will run faster.

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,730 @@
...
+      match item {
+        ModuleItem::ModuleDecl(module_decl) => match module_decl {
+          ModuleDecl::Import(_) => {
+            new_items.push(item.clone());
+          }
+          ModuleDecl::ExportDecl(export_decl) => {
+            if let Some(decl) = self.decl_to_type_decl(&export_decl.decl) {
```
dsherret, line 48: When this does return `None` for some reason, should we consider it an error? Maybe we can at least debug panic?

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,730 @@
...
+      | Decl::TsEnum(_)
+      | Decl::TsModule(_) => Some(decl.clone()),
+      _ => None,
+    }
+  }
+
+  fn class_body_to_type(&mut self, body: &[ClassMember]) -> Vec<ClassMember> {
+    body
+      .iter()
+      .filter_map(|member| match member {
```
dsherret, line 315: Should we handle scenarios where `None` is returned for some reason?

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,730 @@
...
+              }),
+            ))
+          }
+          // Keep all these
+          ModuleDecl::TsImportEquals(_)
+          | ModuleDecl::TsNamespaceExport(_)
+          | ModuleDecl::TsExportAssignment(_) => {
+            new_items.push(ModuleItem::ModuleDecl(module_decl.clone()));
+          }
+          _ => {}
```
dsherret, line 109: Instead of doing a wildcard match (`_ =>`), we can change this to list all `ModuleDecl` variants that will match for this. That way when new a new `ModuleDecl` variant gets added we can get errors here and decide how we should update the code.

Similarly we should do this for all wildcards below. Also, we should probably surface an error for these scenarios (to say we had issues creating a dts file) or debug panic.

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,730 @@
...
+                      computed,
+                      optional: false,
+                      init: None,
+                      params: vec![],
+                      type_ann: init_type,
+                      type_params: None,
+                    },
+                  ));
+                }
+                // TODO: Requires type resolving, skip for now
```
dsherret, line 188: Should we surface a diagnostic here to say we can't properly create a dts file? Also, does this require type resolving? Can we do `ident` -> `ident: typeof ident` or something similar?

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,730 @@
...
+          }
+        }
+
+        Some(ts_readonly(TsType::TsTupleType(TsTupleType {
+          span: DUMMY_SP,
+          elem_types,
+        })))
+      }
+      Expr::Object(obj) => {
+        let mut members: Vec<TsTypeElement> = vec![];
```
dsherret, line 150:
```suggestion
        let mut members: Vec<TsTypeElement> = Vec::with_capacity(obj.len());
```

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,730 @@
...
+                      type_params: None,
+                    },
+                  ));
+                }
+                // TODO: Requires type resolving, skip for now
+                Prop::Shorthand(_) => {}
+                _ => {}
+              }
+            }
+            PropOrSpread::Spread(_) => {}
```
dsherret, line 193: Similarly here?

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,730 @@
...
+            type_ann: return_type,
+            type_params: fn_expr.function.type_params,
+          }),
+        ))
+      }
+      _ => None,
+    }
+  }
+
+  fn decl_to_type_decl(&mut self, decl: &Decl) -> Option<Decl> {
```
dsherret, line 264: We could change this to take an owned value in order to reduce cloning.

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,730 @@
...
+
+    let module = module_info.source().module().clone();
+
+    let mut transformer = FastCheckDtsTransformer::new();
+    let module = transformer.transform(module).unwrap();
+
+    let parsed_source = module_info.source();
+    let comments =
+      CommentsMut::new(parsed_source.comments().as_single_threaded());
+    let comments2 = comments.into_multi_threaded().into_single_threaded();
```
dsherret, line 410: I think we can just do `let comments = parsed_source.comments().as_single_threaded()`?

dsherret (commented):
It's looking good, but we should spend some time handling unknown variants a bit before landing this so that we don't swallow errors and can be alerted better to problems (either by debug panics or surfacing diagnostics that dts generation failed).

src/fast_check/transform_dts.rs
marvinhagemeister: I'm afraid not, it leads to a type mismatch:

```sh
mismatched types
expected reference `&SingleThreadedComments`
   found reference `&CommentsMut`
```

dsherret: What does the full code look like? There shouldn't be a `CommentsMut` in there.

marvinhagemeister: ohh wait you're right, I can drop the whole `CommentsMut` thing. Thanks for the tip 👍

marvinhagemeister:
@dsherret This is ready to be reviewed now. Made an integration into jsr as well.

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,1352 @@
...
+      parsed_source,
+      specifier,
+      diagnostics: vec![],
+    }
+  }
+
+  fn gen_unique_name(&mut self) -> String {
+    self.id_counter += 1;
+    format!("_dts_{}", self.id_counter)
+  }
```
dsherret, line 95: In the future we can make this better to actually avoid conflicts, but the chance of that happening are so slim (ex. do a pre-pass on the file and collect all the identifier names--including globals referenced in types and put them in a hashset, then here it would try something similar, but check for conflicts)

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,1352 @@
...
+              continue;
+            }
+
+            new_items.push(ModuleItem::ModuleDecl(
+              ModuleDecl::ExportDefaultDecl(value),
+            ))
+          }
+          ModuleDecl::ExportDefaultExpr(export_default_expr) => {
+            let is_overload =
+              if let Expr::Fn(fn_expr) = &*export_default_expr.expr.clone() {
```
dsherret, line 209: I think the clone can be removed here?

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,1352 @@
...
+            match decl {
+              Decl::TsEnum(mut ts_enum) => {
+                ts_enum.declare = true;
+                new_items
+                  .push(ModuleItem::Stmt(Stmt::Decl(Decl::TsEnum(ts_enum))));
+              }
+              Decl::TsInterface(_)
+              | Decl::TsTypeAlias(_)
+              | Decl::Using(_)
+              | Decl::TsModule(_) => {
```
dsherret, line 281: Are we handling TypeScript namespaces? Those can contain expressions. Are there any tests for this and for when they're exported from the module?

tests/specs/graph/JsrSpecifiers_FastCheck_TsModule.txt
```diff
@@ -71,3 +71,11 @@ Fast check https://jsr.io/@scope/a/1.0.0/mod.ts:
   }
   class Public1 {
   }
+  --- DTS ---
+  export module Test.Test {
+    export class MyClass extends Public1 {
+      prop!: string;
```
dsherret, line 77: We'll need to remove the `!` here.

tests/specs/graph/JsrSpecifiers_FastCheck_Enums.txt
```diff
@@ -108,3 +108,28 @@ Fast check https://jsr.io/@scope/a/1.0.0/mod.ts:
...
+    Value2 = 2
+  }
+  export declare enum EnumWithStringInits {
+    Value1 = "a",
+    Value2 = "b"
+  }
+  declare const value: number;
+  export declare enum EnumWithNonConstInits {
+    Value1 = new Public1().test,
+    Value2 = new Public2().asdf * value + NonExportedEnum.Value
```
dsherret, line 127: Some expressions here that we'll need to handle somehow (seems like we can just remove them)

dsherret (commented):
Almost there!

tests/specs/graph/JsrSpecifiers_FastCheck_TsModule.txt
marvinhagemeister: Was an symptom of not traversing into TypeScript namespace.

src/fast_check/transform_dts.rs
```diff
@@ -0,0 +1,1509 @@
...
+          | TsType::TsParenthesizedType(_)
+          | TsType::TsTypeOperator(_)
+          | TsType::TsIndexedAccessType(_)
+          | TsType::TsMappedType(_)
+          | TsType::TsTypePredicate(_)
+          | TsType::TsImportType(_) => false,
+        }
+      }
+
+      Expr::TsAs(ts_as) => self.valid_enum_ts_type(*ts_as.type_ann),
```
dsherret, line 731: I get an error for this in declaration files

> In ambient enum declarations member initializer must be constant expression.ts(1066)

tests/specs/graph/JsrSpecifiers_FastCheck_Enums.txt
```diff
@@ -108,3 +120,50 @@ Fast check https://jsr.io/@scope/a/1.0.0/mod.ts:
...
+  declare const NUM: any;
+  export declare enum Foo1 {
+    A = 1,
+    B = "2",
+    C = 1 << 2,
+    D = 1 + 2,
+    F = Foo1.A | Foo1.C,
+    G = Foo1.A + Foo1.C,
+    H,
+    I = 1 + NUM
```
dsherret, line 168: It seems this is also not allowed in dts files.

src/fast_check/transform_dts.rs
marvinhagemeister: This seems to work for a very small subset:
```
export enum Foo {
  A = 1 as number,
}
```

tests/specs/graph/JsrSpecifiers_FastCheck_Enums.txt
marvinhagemeister: Yeah wanted to add a couple cases where TS strips the type but still generates the files even with a type error.

dsherret: We probably shouldn't follow TS here and instead just strip this with a warning diagnostic.

PR #390, opened 2024-02-17
fix: jsr specifier does not support tags

irbull:
When a jsr specifier is used, an error is returned if the version segment contains a tag.

Fixes: https://github.com/denoland/deno/issues/22420

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -39,6 +39,8 @@ use deno_semver::npm::NpmPackageReqReference;
 use deno_semver::package::PackageNv;
 use deno_semver::package::PackageNvReference;
 use deno_semver::package::PackageReq;
+use deno_semver::package::PackageReqReferenceParseError;
+use deno_semver::RangeSetOrTag;
 use deno_semver::Version;
 use futures::future::LocalBoxFuture;
 use futures::future::Shared;
@@ -62,6 +64,7 @@ use std::collections::HashSet;
 use std::collections::VecDeque;
 use std::fmt;
 use std::sync::Arc;
+use thiserror::Error;
 use url::Url;
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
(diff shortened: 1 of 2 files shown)


irbull:
It would be good to create an integration test that shows the error attached to the graph. Any pointers?

src/graph.rs
```diff
@@ -2770,6 +2773,14 @@ struct Builder<'a, 'graph> {
   diagnostics: Vec<BuildDiagnostic>,
 }
 
+#[derive(Error, Debug, Clone)]
+pub enum JsrPackageFormatError {
```
irbull, line 2777: I'm not sure if this is the best way to propagate errors. Suggestions welcome.

src/graph.rs
```diff
@@ -4483,6 +4507,54 @@ mod tests {
...
+    )
+    .is_ok());
+
+    assert!(
+      validate_jsr_specifier(&Url::parse("jsr:@scope/mod").unwrap()).is_ok()
+    );
+  }
+
+  // TODO: Write a test that would demonstrate that jsr does not support version tags
+  //
```
irbull, line 4533: It would be great to have a more involved test or even an integration test. Suggestions welcome.

dsherret:
Make a copy of one of these files:

https://github.com/denoland/deno_graph/blob/main/tests/specs/graph/JsrSpecifiers_VersionNotExist.txt

Then modify and run with `UPDATE=1 cargo test` then inspect the changes it makes.

src/graph.rs
irbull: I've added a test

src/graph.rs
dsherret: This is good thanks! Minor nitpick: would you be able to move it down to above `validate_jsr_specifier` so it's not between `Builder` and its implementation though?

dsherret (approved):
LGTM. Thanks @irbull!

PR #400, opened 2024-02-22
feat: add PackageSpecifiers::used_yanked_packages()

nayeemrmn:


src/graph.rs
```diff
@@ -3426,19 +3426,38 @@ impl<'a, 'graph> Builder<'a, 'graph> {
...
-      if let Some(version) = resolve_version(
-        &package_req.version_req,
-        existing_versions.iter().map(|nv| &nv.version),
-      ) {
-        return Some(version.clone());
-      }
+      existing_versions
+        .iter()
+        .map(|nv| &nv.version)
+        .collect::<Vec<_>>()
```
dsherret, line 3435: I think we should keep the existing code here and prioritize existing versions even if they're yanked because it looks like this would cause the lockfile to not be respected for yanked packages (ex. someone has a package locked, someone yanks a package and publishes a malicious package version, now you're compromised) and also we should prefer to not cause duplicate dependencies by preferring yanked packages that have previously been resolved. The user will get a warning and then they can decide how to proceed.

dsherret (commented):
Looks good, but maybe let's add some tests in the test/specs folder? You can copy one of the txt files there and then change the code, then run with `UPDATE=1 cargo test` to automaticaly update the output of the test.

src/graph.rs
nayeemrmn: But in the lockfile case the code after this will be a filtration of this, which should be a 1-element vec. So if it's yanked, `unyanked_versions` will be empty, right?

nayeemrmn:
How do we represent the lockfile in the test fixture?

dsherret:
@nayeemrmn we might need to add something similar to what's done with workspace members if it doesn't exist: https://github.com/denoland/deno_graph/blob/main/tests/specs/graph/JsrSpecifiers_Workspaces.txt

src/graph.rs
dsherret: Ah, I just misread the code

src/graph.rs
```diff
@@ -3426,19 +3426,38 @@ impl<'a, 'graph> Builder<'a, 'graph> {
...
+    } else {
+      package_info.versions.keys().collect::<Vec<_>>()
+    };
+    let unyanked_versions = versions.clone().into_iter().filter(|v| {
```
dsherret, line 3439: Is cloning the versions necessary here? I'm guessing necessary because of the borrow checker somehow.

src/graph.rs
```diff
@@ -3426,19 +3426,38 @@ impl<'a, 'graph> Builder<'a, 'graph> {
...
+    if let Some(version) =
+      resolve_version(&package_req.version_req, versions.into_iter()).cloned()
+    {
+      self.graph.packages.add_used_yanked_package(PackageNv {
+        name: package_req.name.clone(),
+        version: version.clone(),
+      });
+      return Some(version);
+    }
+    None
```
dsherret, line 3460: I'm having trouble reading this code. Can we change it to be a bit more explicit in its steps?

For example, it might be nice if the code read like:

1. Check the resolved versions.
  a. Is it yanked? Add to yanked. Done.
2. Resolve with the unyanked versions.
3. Resolve including the yanked versions.
  a. Add to yanked. Done.

src/graph.rs
```diff
@@ -3426,19 +3426,38 @@ impl<'a, 'graph> Builder<'a, 'graph> {
...
+        .collect::<Vec<_>>()
+    } else {
+      package_info.versions.keys().collect::<Vec<_>>()
+    };
+    let unyanked_versions = versions.clone().into_iter().filter(|v| {
+      package_info
+        .versions
+        .get(v)
+        .map(|i| !i.yanked)
+        .unwrap_or(true)
```
dsherret, line 3444: We could avoid reaching into the hashmap for every version by instead only doing this if the version is resolved. See my comment below.

tests/helpers/mod.rs
```diff
@@ -246,26 +256,28 @@ fn parse_spec(text: String) -> Spec {
   files.push(current_file.unwrap());
   let output_file =
     files.remove(files.iter().position(|f| f.specifier == "output").unwrap());
-  let diagnostics = take_vec_file(&mut files, "diagnostics");
-  let workspace_members = take_vec_file(&mut files, "workspace_members");
+  let diagnostics = take_file(&mut files, "diagnostics");
+  let workspace_members = take_file(&mut files, "workspace_members");
+  let lockfile_jsr_packages = take_file(&mut files, "lockfile_jsr_packages");
```
dsherret, line 261: Nice! I think you still have the test that uses this locally though? Or I might be missing it.

dsherret (commented):
Almost there!

src/graph.rs
nayeemrmn: > if the version is resolved

Do you mean unresolved? It won't catch lockfile entries then, since they are in `existing_versions` but not yielded from this function.

src/graph.rs
nayeemrmn: It's just a vec of refs, yeah it's to avoid a double ref on the versions in the iterator. Could be done by mapping as well.

nayeemrmn: Changed to `.iter().copied()`.

src/graph.rs
```diff
@@ -3426,19 +3426,46 @@ impl<'a, 'graph> Builder<'a, 'graph> {
...
+    }
+    if let Some(version) = resolved_yanked_version {
+      self.graph.packages.add_used_yanked_package(PackageNv {
+        name: package_req.name.clone(),
+        version: version.clone(),
+      });
+      return Some(version);
+    };
+    None
   }
```
dsherret, line 3469: I think we could reduce the hashmap lookups and cloning by doing something like this?

```rs
  fn resolve_jsr_version(
    &mut self,
    package_info: &JsrPackageInfo,
    package_req: &PackageReq,
  ) -> Option<Version> {
    // 1. try to resolve with the list of existing versions
    if let Some(existing_versions) =
      self.graph.packages.versions_by_name(&package_req.name)
    {
      if let Some(version) = resolve_version(
        &package_req.version_req,
        existing_versions.iter().map(|nv| &nv.version),
      ) {
        let is_yanked = package_info
          .versions
          .get(version)
          .map(|i| i.yanked)
          .unwrap_or(false);
        let version = version.clone();
        if is_yanked {
          self.graph.packages.add_used_yanked_package(PackageNv {
            name: package_req.name.clone(),
            version: version.clone(),
          });
        }
        return Some(version);
      }
    }

    // 2. attempt to resolve with the unyanked versions
    let unyanked_versions =
      package_info.versions.iter().filter_map(|(v, i)| {
        if !i.yanked {
          Some(v)
        } else {
          None
        }
      });
    if let Some(version) =
      resolve_version(&package_req.version_req, unyanked_versions)
    {
      return Some(version.clone());
    }

    // 3. attempt to resolve with the the yanked versions
    let yanked_versions = package_info.versions.iter().filter_map(|(v, i)| {
      if i.yanked {
        Some(v)
      } else {
        None
      }
    });
    if let Some(version) =
      resolve_version(&package_req.version_req, yanked_versions)
    {
      self.graph.packages.add_used_yanked_package(PackageNv {
        name: package_req.name.clone(),
        version: version.clone(),
      });
      return Some(version.clone());
    }

    None
  }
```

src/packages.rs
```diff
@@ -20,8 +20,14 @@ pub struct JsrPackageInfo {
...
 
+fn is_false(v: &bool) -> bool {
+  !v
+}
+
 #[derive(Debug, Serialize, Deserialize, Clone, Default)]
 pub struct JsrPackageInfoVersion {
+  #[serde(default, skip_serializing_if = "is_false")]
+  pub yanked: bool,
   // no used fields yet
```
dsherret, line 31: Oh, we can remove this comment now.

dsherret (approved):
LGTM! Thanks @nayeemrmn!

Have you tried integrating this one into the CLI yet? It might be good to do that before merging in case we missed something, but I don't think we did.

src/packages.rs
nayeemrmn, line 31:
```suggestion
```

PR #411, opened 2024-03-11
feat(fast_check): infer type literals

marvinhagemeister:
This PR adds support for inferring the type of an expression initializer for an exported var declaration. This is needed to make branded types work with fast-check and our dts transform.

```ts
// input
export const foo = "foo" as "foo" & { __brand: "foo" };

// Fast check output
export const foo: "foo" & { __brand: "foo" } = {} as any;

// DTS
export const foo: "foo" & { __brand: "foo" };
```

### DIFF
--- a/src/fast_check/transform.rs
+++ b/src/fast_check/transform.rs
@@ -1568,7 +1568,7 @@ fn infer_simple_type_from_type(t: &TsType) -> Option<TsType> {
       },
     })),
     TsType::TsTypeQuery(_) => None,
-    TsType::TsTypeLit(_) => None,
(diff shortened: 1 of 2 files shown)


dsherret (approved):
LGTM!

dsherret (changes requested):
Can you add a test to this PR that does something like?

```
class Public {}
export const str = "foo" as "foo" & { data: Public };
```

`Public` should show up in the output.

dsherret (approved):
Investigated and added assertion because I want to do a deno_graph release.

PR #418, opened 2024-03-17
feat: @deno-types map in build options

nayeemrmn:
Adds support for a `deno.json` field like:
```json
{
  "types": {
    "npm:express": "npm:@types/express",
    "./file.js": "./file.d.ts"
  }
}
```
All specifiers written there are resolved through provided resolver with the `deno.json` as the referrer. So an import-mapped key (or value) will just work:
```json
{
  "imports": {
    "express": "npm:express@^4.18.3"
  },
  "types": {
    "express": "npm:@types/express@^4.17.21"
  }
}
```
We can make it so that `deno add npm:@types/<package>` adds the entry to `"types"` instead of `"imports"`. So `deno add npm:@types/express` will be the equivalent of `npm i --save @types/express`.

These mappings will only apply to referrers within the passed 'scope' associated with the `deno.json`. In CLI we would presumably pass the parent directory of the `deno.json` file.

JSR will inline these mappings to pragmas on publish. This has parity with the above behaviour.

---

This is essentially `// @deno-types` pragmas in config file form. Reasons we want to do this:
- Deduplicate `// @deno-types` pragmas for dependencies.
- Pragmas can't be used for things like `compilerOptions.jsxImportSource`.
- NPM `@types` packages can be referenced and version-pinned in the config file next to their runtime counterparts.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4104,10 +4104,11 @@ export function a(a: A): B {
         !matches!(e, ModuleGraphError::ModuleError(ModuleError::Missing(..)))
       })
       .collect::<Vec<_>>();
-    assert_eq!(errors.len(), 3);
-    assert_eq!(errors[0].to_string_with_range(), "Relative import path \"bad_key\" not prefixed with / or ./ or ../\n    at file:///a/deno.json:1:1");
-    assert_eq!(errors[1].to_string_with_range(), "Relative import path \"bad_value\" not prefixed with / or ./ or ../\n    at file:///a/deno.json:1:1");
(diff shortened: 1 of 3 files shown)


marvinhagemeister:
I quite like this idea. Makes a lot of sense to put it in the config file for imports defined there 👍

src/lib.rs
```diff
@@ -3969,6 +3983,121 @@ export function a(a: A): B {
     }
   }
 
+  #[tokio::test]
```
dsherret, line 3986: Would be good to add a test for where there's a `type` entry and an import has a `deno-types` pragma (the pragma should win).

PR #422, opened 2024-03-25
feat: infer new Map/Set expression type

marvinhagemeister:
This PR adds support for inferring the type of `new Map/Set` expressions if the relevant type arguments are specified:

```ts
export const foo = new Map<string, string>();
export const bar = new Set<string>();
export function something(arg = new Map<string, string>(), arg2 = new Set<string>()): void {}
```

Inferred result:

```ts
export declare const foo: Map<string, string>;
export declare const bar: Set<string>;
export declare function something(arg?: Map<string, string>, arg2?: Set<string>): void;
```

Fixes https://github.com/denoland/deno/issues/23068

### DIFF
--- a/tests/specs/graph/jsr/FastCheck_Infer_Map_Set.txt
+++ b/tests/specs/graph/jsr/FastCheck_Infer_Map_Set.txt
@@ -1,0 +1,78 @@
+# https://jsr.io/@scope/a/meta.json
+{"versions": { "1.0.0": {} } }
+
+# https://jsr.io/@scope/a/1.0.0_meta.json
+{ "exports": { ".": "./mod.ts" } }
+
+# https://jsr.io/@scope/a/1.0.0/mod.ts
+type Foo = { foo: number }
+export const foo = new Map<string, Foo>();
+export const bar = new Set<Foo>();
+
+export function something(arg = new Map<string, string>(), arg2 = new Set<string>()): void {
+
+}
+
+# mod.ts
+import 'jsr:@scope/a'
+
+# output
+{
+  "roots": [
+    "file:///mod.ts"
+  ],
+  "modules": [
+    {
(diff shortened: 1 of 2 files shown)


tests/specs/graph/jsr/FastCheck_Infer_Map_Set.txt
```diff
@@ -0,0 +1,78 @@
...
+  "packages": {
+    "@scope/a": "@scope/a@1.0.0"
+  }
+}
+
+Fast check https://jsr.io/@scope/a/1.0.0/mod.ts:
+  {}
+  type Foo = {
+    foo: number;
+  };
```
dsherret, line 68: I'm surprised it included this type because there was no changes in the public type analysis. I wonder if we are too eagerly grabbing out types or if there's something about how the analysis works that I've forgotten.

dsherret (commented):
I'm not sure we should go down this path because it's a bit of a slippery slope. Certain libraries might require overriding globals and then the globals could use inference in a way, that causes fast check to break down.

That said, I think this is ok in this scenario with general collection types, but we need to be super strict in what we allow (ex. zero arguments provided in the constructor).

marvinhagemeister:
@dsherret good point about checking global symbol identity + number of arguments. I think inference of both the `Map` and `Set` type is acceptable and similar to inferring literals.

lucacasonato:
Mh, I am also a bit weary of this. I agree with David that we need to be very careful about things like this.

It seems a bit confusing we'd support this for `Map` and `Set`, but not for any other interfaces that have generics, where the user specified concrete types for all of the generics (because I think that is the rule here). I do not think we can generalize this however, because that would require cross module inference.

marvinhagemeister:
No worries that's fine. In that case it sounds like it's best to close this PR.

PR #428, opened 2024-04-08
fix(fast_check): support expando properties

marvinhagemeister:
This PR addresses an issue with FastCheck where the following snippet would lead to a type error:

```ts
export interface it {
  skip(): void
}
export function it() {}
it.skip = () => {}

it.skip()
```

Internally, TypeScript generates a namespace for `it` which tells the compiler that the function has a `skip` property. The `inteface` itself is interestingly ignored. This lead to us erroring.

The solution is to generate the `namespace` declaration which we need to anyway for the dts transform.

```ts
export interface it {
    skip(): void;
}
export function it(): void {}
export namespace it {
    var skip: () => void;
}
```

This matches the exact output that `tsc` generates.

Fixes https://github.com/denoland/deno/issues/23276


marvinhagemeister:
This fixes the issue, but it likely needs further iteration. Especially in regards to `maybe_infer_type_from_expr` which I feel like should return a `Return` type. What do you think @dsherret ?

src/fast_check/transform.rs
dsherret, line 1: We should add this crazy test case and make it work:

```ts
export function test() {
}

if (true) {
    test.skip = () => {};
}

test.skip();

export namespace Test {
    export function test() {
    }

    test.other = () => {};

    test.other();
}

export namespace Test {
    test.other2 = () => {};

    test.other2();
}
```

This is TypeScript's dts output:

```ts
export declare function test(): void;
export declare namespace test {
    var skip: () => void;
}
export declare namespace Test {
    function test(): void;
    namespace test {
        var other: () => void;
        var other2: () => void;
    }
}
export declare namespace Test {
}
```

tests/specs/graph/jsr/FastCheck_Vars.txt
```diff
@@ -148,16 +148,14 @@ Fast check https://jsr.io/@scope/a/1.0.0/mod.ts:
...
-  export const func7 = (): Promise<1> =>({} as never);
-  export const func8 = async ()=>"test" as const;
+  export const func1: () => string = {} as never;
+  export const func2: () => string = {} as never;
+  export const func3: () => string = {} as never;
+  export const func4: () => "test" = {} as never;
+  export const func5: () => Promise<string> = {} as never;
+  export const func6: () => Promise<1> = {} as never;
+  export const func7: () => Promise<1> = {} as never;
+  export const func8: () => Promise<"test"> = {} as never;
```
dsherret, line 158: As part of this change, let's revert this and get the output more like before in order to focus the change a bit more. I need to spend some time investigating the reason why we don't do this OOTB and to document it. I remember there was a good reason (maybe not for this specific case), but I'm pre-occupied at the moment to investigate (will get to it soon). Then we can follow up later with this change in a separate PR.

src/fast_check/transform.rs
```diff
@@ -1821,6 +2368,15 @@ fn is_expr_ident_or_member_idents(expr: &Expr) -> bool {
...
 }
 
+fn contains_return_stmt(stmts: &Vec<Stmt>) -> bool {
+  for stmt in stmts {
+    if let Stmt::Return(_) = stmt {
+      return true;
+    }
+  }
+  false
+}
```
dsherret, line 2378: This will return false for:

```ts
if (true) {
  return true;
} else {
  return false;
}
```

We can use `get_return_stmts_with_arg_from_stmts`, which already exists to figure this out.

src/fast_check/transform.rs
```diff
@@ -1463,27 +1668,260 @@ impl<'a> FastCheckTransformer<'a> {
     Ok(is_leavable)
   }
 
-  fn maybe_infer_type_from_expr(&self, expr: &Expr) -> Option<TsType> {
+  fn maybe_infer_type_from_expr(
+    &mut self,
+    expr: &Expr,
+    report_diagnostics: bool,
+  ) -> Result<Option<TsType>, Vec<FastCheckDiagnostic>> {
```
dsherret, line 1675: Can we separate out the inference into a separate PR (or it looks like multiple separate PRs)? Then we can really make sure we get a good end result for each of them. Doing inference is kind of risky with fast check.

src/fast_check/transform.rs
```diff
@@ -185,6 +186,7 @@ struct FastCheckTransformer<'a> {
   parsed_source: &'a ParsedSource,
   should_error_on_first_diagnostic: bool,
   diagnostics: Vec<FastCheckDiagnostic>,
+  public_inferred_namespaces: HashMap<String, Vec<ModuleItem>>,
```
dsherret, line 189: Right now this won't handle stuff like namespaces. Maybe this should instead be a stack of hashmaps where the top item represents the current module or namespace and then when we pop from the stack we add additional items to the namespace items.

src/fast_check/transform.rs
```diff
@@ -229,7 +232,7 @@ impl<'a> FastCheckTransformer<'a> {
     comments: &mut CommentsMut,
     is_ambient: bool,
   ) -> Result<Vec<ModuleItem>, Vec<FastCheckDiagnostic>> {
-    let mut final_body = Vec::with_capacity(body.len());
+    let mut final_body = vec![];
```
dsherret, line 235: Maybe `body.len() + self.public_inferred_namespaces.iter().map(|v| v.len()).count()`.

src/fast_check/transform.rs
```diff
@@ -442,6 +509,11 @@ impl<'a> FastCheckTransformer<'a> {
         if !self.public_ranges.contains(&public_range) {
           return Ok(false);
         }
+
+        self
+          .public_inferred_namespaces
+          .insert(n.ident.sym.to_string(), vec![]);
```
dsherret, line 515: Instead of a string, maybe let's store `n.ident.sym` in the hashmap because that takes scope into account too. Here's a test case that shouldn't have a `skip` property:

```ts
export function test() {
}

if (true) {
  function test() {}
  test.skip = () => {};
}
```

dsherret (commented):
Looks great, but we need to make this work for namespaces.

src/fast_check/transform.rs
dsherret: Nope. This is a bad suggestion because the public_inferred_namespaces will be empty here.

src/fast_check/transform.rs
dsherret: We'll save this one for later.

tests/specs/symbols/expando_properties.txt
```diff
@@ -0,0 +1,267 @@
...
+                    flags: 0,
+                },
+            ],
+            child_ids: {
+                2,
+                3,
+            },
+            exports: {
+                "skip": 2,
+                "prop": 3,
```
dsherret, line 104: See here how the `myFunc` symbol now has these expando properties as exports.

tests/specs/ecosystem/azmi/kitchensink/0_7_5.test
```diff
@@ -1435,17 +1435,6 @@ error[missing-explicit-type]: missing explicit type in the public API
   docs: https://jsr.io/go/slow-type-missing-explicit-type
 
 
-error[missing-explicit-type]: missing explicit type in the public API
-   --> /src/binder.ts:270:2
-    | 
-270 |   bind_array_clear = /*@__PURE__*/ bindMethodFactoryByName(array_proto, "splice", 0) as <T>(array: T[]) => ((deleteCount?: number, ...items: T[]) => T[]),
```
dsherret, line 1441: It now changes this to:

```
bind_array_clear = ({} as never) as <T>(array: T[]) => ((deleteCount?: number, ...items: T[]) => T[]),
```

This was added in this PR in order to support defining a type for expando properties via an assertion.

tests/specs/ecosystem/core/match/0_1_5.test
```diff
@@ -18,6 +18,15 @@ core/match/0.1.5
 
 ===
 
-== FAST CHECK EMIT PASSED ==
+== FAST CHECK EMIT FAILED ==
+error[unsupported-expando-property]: expando property referencing 'greedy' conflicts with 'placeholder.greedy'
+   --> /mod.ts:107:22
+    | 
+107 | placeholder.greedy = greedy;
```
dsherret, line 25: We can't transform:

```ts
export function test() {
}

test.greedy = greedy;

export function greedy() {
}
```

To:

```ts
export function test() {
}

export namespace test {
  export var greedy = greedy; // notice issue here
}

export function greedy() {
}
```

...so we just raise a diagnostic.

src/fast_check/transform.rs
```diff
@@ -390,8 +493,48 @@ impl<'a> FastCheckTransformer<'a> {
         | Stmt::DoWhile(_)
         | Stmt::For(_)
         | Stmt::ForIn(_)
-        | Stmt::ForOf(_)
-        | Stmt::Expr(_) => Ok(false),
+        | Stmt::ForOf(_) => Ok(false),
+        Stmt::Expr(n) => match &mut *n.expr {
```
lucacasonato, line 497:
```suggestion
        Stmt::Expr(Expr::Assign(assign_expr)) => {
```

And then `Stmt::Expr(_) => Ok(false)` for the rest.

Can you also move the `Ok(false)` branch to the end of the `match`?

tests/specs/ecosystem/std/expect/0_224_0.test
```diff
@@ -7,6 +7,202 @@ std/expect/0.224.0
 
 ===
 
-== FAST CHECK EMIT PASSED ==
+== FAST CHECK EMIT FAILED ==
+error[unsupported-expando-property]: expando property referencing 'anything' conflicts with 'expect.anything'
+   --> /expect.ts:209:19
+    | 
+209 | expect.anything = anything;
```
lucacasonato, line 14: Seems quite common - oh well, I guess we can address in a follow up.

tests/specs/ecosystem/core/match/0_1_5.test
lucacasonato: There seems to be quite a few examples of this in the ecosystem tests - oh well we can address in a follow up.

lucacasonato (approved):
LGTM - how complicated!

PR #431, opened 2024-04-10
feat: `follow_fast_check_modules` option for conditional walking of fast check modules

iuioiua:
Previously, when a slow type was added to a file, the fast check was skipped, and all specifiers were checked. This new option, if toggled, allows you to check all specifiers anyway.

We should add a test, but I'm not sure how to.

Towards fixing https://github.com/denoland/deno/issues/23023


dsherret (approved):
Thanks! LGTM. This is kind of hard to test, but we can add a test in the CLI for it.

PR #432, opened 2024-04-11
fix(fast_check): properly handle rest parameters in overloads

lucacasonato:
Fixes https://github.com/jsr-io/jsr/issues/376

### DIFF
--- a/tests/specs/graph/jsr/FastCheck_Overloads.txt
+++ b/tests/specs/graph/jsr/FastCheck_Overloads.txt
@@ -17,6 +17,12 @@ export function overloadSpread(...args: unknown[]) {
   return args;
 }
 
+export function overloadLast(value: string, ...args: string[]);
+export function overloadLast(value: number, ...args: number[]);
+export function overloadLast(value: string | number, ...args: unknown[]) {
+  return args;
+}
+
 # mod.ts
 import 'jsr:@scope/a'
 
@@ -52,7 +58,7 @@ import 'jsr:@scope/a'
     },
     {
       "kind": "esm",
-      "size": 328,
+      "size": 549,
       "mediaType": "TypeScript",
       "specifier": "https://jsr.io/@scope/a/1.0.0/mod.ts"
     }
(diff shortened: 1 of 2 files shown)


tests/specs/graph/jsr/FastCheck_Overloads.txt
```diff
@@ -0,0 +1,89 @@
...
+  argsToArray(..._args: string[]): void;
+  argsToArray(..._args: number[]): void;
+  argsToArray(..._args: unknown[]): void {}
+}
+
+function overloadSpread(...args: string[]);
+function overloadSpread(...args: number[]);
+export function overloadSpread(...args: unknown[]) {
+  return args;
+}
```
marvinhagemeister, line 18: Mixing non-exported overloads with an exported implementation is not valid in TS. See https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwDYEMDOa4qgczQFQgEEooUBPOAbwCg4tcDjSKAKAOk4H1s8AuOGhhQAlgDscAbQC6ASgEA3CCIAmAbjoM8hEmXIduvNALEBXALYAjYFBny4S1RvpGdLfZ3Y9GA02IDWYhAA7mJ2isoq1AC+NLE0AGZ+CDAiEGJwEAo2SBAoKgDKYFDA+QbsRgJCohLhWGLkGkliKWkZWTl5hcWlKuWVcGZWNnUoDRqgkLBwza3pmdlQuflFJWWeA36BIWFyAmOUtPQlMKZQGUYa0UA

tests/specs/graph/jsr/FastCheck_Overloads.txt
```diff
@@ -0,0 +1,89 @@
...
+    return {} as never;
+  }
+  --- DTS ---
+  export declare class argsToArray {
+    argsToArray(..._args: string[]): void;
+    argsToArray(..._args: number[]): void;
+  }
+  declare function overloadSpread(...args: string[]);
+  declare function overloadSpread(...args: number[]);
+  export declare function overloadSpread(...param0?: any): any;
```
marvinhagemeister, line 89: The implementation is not included in the dts output from `tsc` as it cannot be referenced. Might have to do with the overloads not being exported.

tests/specs/graph/jsr/FastCheck_Overloads.txt
lucacasonato: whoops - you are right. fixed

tests/specs/graph/jsr/FastCheck_Overloads.txt
lucacasonato: yup that was it, fixed.

marvinhagemeister (approved):

atjn:
Wow that was a fast turnaround. Thank you very much!

dsherret (approved):
LGTM

PR #444, opened 2024-04-16
feat(unstable): add @ts-types and clean up @deno-types

lucacasonato:
This allows a `@ts-types` pragma to be used
instead of the `@deno-types` pragma. The
`@ts-types` pragma does not allow unquoted
specifiers.

This also removes the `leadingComment` field in
the `DependencyDescriptor` struct in favor of a
`typesSpecifier` field. This requires bumping the
`ModuleInfo` version to 2.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -2647,12 +2647,11 @@ fn fill_module_dependencies(
             maybe_npm_resolver,
           )
         } else {
-          let range = import.range.clone();
           // only check if the code resolution is for the same range
-          if Some(&range) == dep.maybe_code.maybe_range() {
+          if Some(&import.range) == dep.maybe_code.maybe_range() {
             let types_resolution = resolve(
               &import.specifier,
-              range,
+              import.range.clone(),
               ResolutionMode::Types,
               jsr_url_provider,
               maybe_resolver,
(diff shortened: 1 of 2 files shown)


src/packages.rs
```diff
@@ -85,9 +88,18 @@ impl JsrPackageVersionInfo {
...
-    serde_json::from_value(module_info.clone()).ok()
+    if let Some(module_graph) = self.module_graph_2.as_ref() {
+      let module_graph = module_graph.as_object()?;
+      let module_info = module_graph.get(specifier)?;
+      serde_json::from_value(module_info.clone()).ok()
+    } else if let Some(module_graph) = self.module_graph_1.as_ref() {
+      let module_graph = module_graph.as_object()?;
+      let mut module_info = module_graph.get(specifier)?.clone();
+      module_graph_1_to_2(&mut module_info);
+      serde_json::from_value(module_info.clone()).ok()
```
dsherret, line 99:
```suggestion
      serde_json::from_value(module_info).ok()
```

dsherret (approved):
LGTM!

src/graph.rs
```diff
@@ -2576,40 +2575,38 @@ fn fill_module_dependencies(
...
+        let maybe_type = if let Some(deno_types) = &deno_types {
+          resolve(
+            &deno_types.text,
+            Range::from_position_range(specifier, deno_types.range),
+            ResolutionMode::Types,
+            maybe_resolver,
+            maybe_npm_resolver,
+          )
+        } else {
+          let range = import.range.clone();
```
dsherret, line 2587: This code was here before, but we can defer this clone.

PR #445, opened 2024-04-16
feat(unstable): add support for @ts-self-types pragma

### LINKED ISSUE #442: Support `// @deno-self-types="..."` as an alternative to `/// <reference types="..." />`
As the title says.

lucacasonato:
This commit adds a `@ts-self-types="..."` pragma that
may be placed at the top of untyped code files,
such as JS and JSX files. The specifier in this
pragma will then be used as the types for this
file when it is imported into a different file.

This is effectively the same as the existing
`/// <reference types="..." />` support in untyped
files, but with the added benefit of not
overloading the TypeScript triple-slash directive
syntax, and working with block comments.

Closes #442

### DIFF
--- a/src/fast_check/range_finder.rs
+++ b/src/fast_check/range_finder.rs
@@ -8,7 +8,6 @@ use std::collections::VecDeque;
 use std::sync::Arc;
 
 use deno_ast::swc::ast::Expr;
-use deno_ast::MediaType;
 use deno_ast::SourceRange;
 use deno_ast::SourceRangedForSpanned;
 use deno_semver::package::PackageNv;
@@ -1212,7 +1211,7 @@ impl<'a> PublicRangeFinder<'a> {
 fn is_module_typed(module: &crate::Module) -> bool {
   match module {
     crate::Module::Js(m) => {
-      is_typed_media_type(m.media_type) || m.maybe_types_dependency.is_some()
+      m.media_type.is_typed() || m.maybe_types_dependency.is_some()
     }
     crate::Module::Json(_) => true,
     crate::Module::Npm(_)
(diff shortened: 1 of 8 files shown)


src/graph.rs
```diff
@@ -2648,7 +2669,7 @@ fn analyze_dynamic_arg_template_parts(
 
 /// Determine if a media type is "untyped" and should be checked to see if there
 /// are types provided.
-fn is_untyped(media_type: &MediaType) -> bool {
+pub fn is_untyped(media_type: &MediaType) -> bool {
```
dsherret, line 2672: Maybe we should change this to something like `is_untyped_script` or something like that (or switch to `is_typed_media_type`). Btw, `MediaType` is copy so we can change this to just `MediaType`.

src/fast_check/range_finder.rs
```diff
@@ -1212,7 +1212,7 @@ impl<'a> PublicRangeFinder<'a> {
 fn is_module_typed(module: &crate::Module) -> bool {
   match module {
     crate::Module::Js(m) => {
-      is_typed_media_type(m.media_type) || m.maybe_types_dependency.is_some()
+      !is_untyped(&m.media_type) || m.maybe_types_dependency.is_some()
```
dsherret, line 1215: Maybe we should switch to just using `is_typed_media_type` instead of `is_untyped` so that we don't have double negatives?

src/graph.rs
```diff
@@ -2079,8 +2079,25 @@ pub(crate) fn parse_js_module_from_module_info(
   let mut module = JsModule::new(specifier.clone(), source);
   module.media_type = media_type;
 
-  // Analyze the TypeScript triple-slash references
+  // Analyze the TypeScript triple-slash references and self types specifier
   if graph_kind.include_types() {
+    if is_untyped(&module.media_type) {
```
dsherret, line 2084: Maybe we should get rid of this condition since `module_info.self_types_specifier` will never be set for unsupported media types?

dsherret (approved):
LGTM, but it's unfortunate to add another "deno" specific pragma when this will be published on JSR for everyone to use. I wonder if we should just do `@ts-self-types` to make it generic as it's likely not to conflict with TypeScript since TypeScript will probably not implement this kind of pragma.

dsherret (changes requested):
Since we're coming up with something new that already exists, let's hold off on merging until we can come up with a name that doesn't have deno in the title (and for deno-types)

dsherret (approved):
LGTM

dsherret:
Let's merge this one as undocumented for the time being and subject to change.

PR #448, opened 2024-04-17
feat: add support for @jsxImportSourceTypes pragma

### LINKED ISSUE #443: Add support for `// @jsxImportSourceTypes="..."` and `jsxImportSourceTypes` options
These would allow you to explicitly specify types for the JSX import source file.

It is the logical equivalent of `// @deno-types="..."` ontop of the emitted JSX import.

lucacasonato:
Closes #443


src/graph.rs
```diff
@@ -2167,6 +2167,19 @@ pub(crate) fn parse_js_module_from_module_info(
           })
       })
     });
+    let types_res = module_info.jsx_import_source_types.or_else(|| {
```
lucacasonato, line 2170: Actually I think this behaviour is wrong. We should only use the `default_jsx_import_source_types` for the `default_jsx_import_source`, but never for a `@jsxImportSource` pragma. However `@jsxImportSourceTypes` should be allowed even if there is no `@jsxImportSource` pragma and we default `default_jsx_import_source`.

@dsherret WDYT?

lib/lib.rs
```diff
@@ -216,12 +224,14 @@ pub async fn js_create_graph(
       .map_err(|err| JsValue::from(js_sys::Error::new(&err.to_string())))?;
   let mut loader = JsLoader::new(load, maybe_cache_info);
   let maybe_resolver = if maybe_default_jsx_import_source.is_some()
+    || maybe_jsx_import_source_module.is_some()
     || maybe_jsx_import_source_module.is_some()
```
dsherret, line 228: This is the same as the above.

src/graph.rs
```diff
@@ -2190,6 +2203,25 @@ pub(crate) fn parse_js_module_from_module_info(
           maybe_npm_resolver,
         );
       }
+      if graph_kind.include_types() && dep.maybe_type.is_none() {
+        if let Some(import_source_types) = types_res {
```
dsherret, line 2207: Looks like constructing `types_res` could be conditionally done in here?

src/lib.rs
```diff
@@ -1109,6 +1109,125 @@ console.log(a);
     );
   }
 
+  #[tokio::test]
+  async fn test_build_graph_jsx_import_source_types() {
```
dsherret, line 1113: Looks good. In case you want to add more tests for this, doing it and expanding on the functionality in `tests/specs/graph` is probably easier than manually setting up a loader and such (there is a concept of "spec options" where adding compiler options could probably be done).

src/graph.rs
dsherret: Yeah, I think so, but would you be able to expand on the reasoning why it should behave that way? (Maybe as a comment in the source code)

dsherret (approved):
LGTM once the couple of issues are resolved.

src/graph.rs
lucacasonato: Added reasoning and updated this code to match my description.

dsherret (approved):

PR #450, opened 2024-04-18
feat: add Dependency::with_new_resolver()

nayeemrmn:
This can be used in `Document::maybe_with_new_resolver()` to recompute resolutions without using `Document::maybe_parsed_source`.

We shouldn't be holding references to parsed/raw sources for non-open documents in memory. This would eliminate one usage.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -628,12 +628,15 @@ pub struct Import {
 }
 
 #[derive(Debug, Default, Clone, Serialize)]
+#[cfg_attr(test, derive(Eq, PartialEq))]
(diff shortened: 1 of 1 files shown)


src/graph.rs
```diff
@@ -628,12 +628,15 @@ pub struct Import {
 }
 
 #[derive(Debug, Default, Clone, Serialize)]
+#[cfg_attr(test, derive(Eq, PartialEq))]
```
dsherret, line 631: Maybe we should just add this unconditionally so we can also potentially use it in the cli?

dsherret (approved):
Great idea!

PR #464, opened 2024-05-02
tests: add an ecosystem check

lucacasonato:
This commit adds a test that can run against all published packages on JSR to check whether their fast check behavior has changed.


dsherret (approved):

dsherret:
Amazing!

Left some comments in a dm because my browser is struggling on this pr.

PR #470, opened 2024-05-13
fix: handle return statements in functions correctly

### LINKED ISSUE #468: Update when we can infer `void` from a function body
Right now we sometimes incorrectly infer the function return type due to our `return` statement analysis. The only cases where we can safely infer `void` is when:

- we are in a function declaration, there are no returns with value
- or when we are in a function expression or arrow function expression and either:
  - a single return statement without value is anywhere in the body
  - the body is empty

Notably, we can not infer a function expression from expressions such as `() => { n() }`, because the return type may be `never` if `n()` returns `never`.

Sync functions: https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXxAA8sMAKASgC55UQA3EGAbgCgWiAHHGDJNTXPigV4AbxbxJBYmXKsAvm07deKdNjzwARiPFT4cDMhioWi9oS48+6wfDC6JUrInikA5hBxaoEACoAFlgAzuRiTvqSRCQUrPqK+obGpubK1moCmsCO+mZKlir2eMG8CAC8bmFlAHzh+tGyCvlWvGDFqvAVIjV1UkkmeRYtRagl8O6dlZ21es6uHl4+-kGhvZHSMXIR8Al9IEYDqQXWbaO8AZPdM9uDaa3t8FiTGRqoOVINsbfH92fwAFbPfivd6SfopZqFU5jADWQNseFBj3mnm8vkCITCs3Wny28W24O+w2hvAg8MybyxN3kQA
Async functions: https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXxAA8sMAKASgC55UQA3EGAbgCgWiAHHGDeKAZwCe6JGky58UCvADeLeAoLEy5VgF82nbrwHCwo9NjzwARtLmL4cDMhioWG9oS48+QkSkMT4Yc-MVYiPCkAOYQOCZQEAAqABZY-OSy-pYKRCQUrJYalta29o5arroeYkb4wH6WDprO2j54-LwIALxuesFJLQB8yZbpKuq1LrxgjbxBbSX60j19inl2NU4jDahN8CHwU+4zXb0WAUGh4ZEx8YnzqUoZqinwOQsgNkuFda5j67yx2+0iswd7ssiqNxvAsL9pgZxHgqooBplge9QV94AArSG7aHlOEKRYFYb1T4bADWmI6nhhqFx4OOYQiUTiCSSh2uCLu2Xu+KRq2JvAg5NKXlhLLYaiAA

lucacasonato:
We now correctly make a distinction between expression like and declaration like functions when inferring return types based on return statements.

The biggest change here is that expression-like functions that have no `return` statements in the function body will now not infer to `void` anymore. This is because in some cases TypeScript will instead infer the return type to `never`. This applies to function expressions, arrow functions, and object literal function expression shorthand. It does not apply to function declarations and class members.

Fixes #468


marvinhagemeister (commented):
LGTM code wise! I'm admittedly a bit unsure of the implications from product a product perspective. We seem to be able to very accurately infer whether a function should return `void` to the point to where we could add the missing return type ourselves, rather than asking the user to do that.

lucacasonato:
> LGTM code wise! I'm admittedly a bit unsure of the implications from product a product perspective. We seem to be able to very accurately infer whether a function should return `void` to the point to where we could add the missing return type ourselves, rather than asking the user to do that.

In all the cases we are 100% sure, we do this inference already. However in the cases that we now disallow, TypeScript can infer both to `void` and `never` based on type information.

For example this code block infers never as the return type:

```ts
() => {
  Deno.exit();
}
```

But this example infers `void`:

```ts
() => {
  fetch()
}
```

We cannot make this distinction without type info, so we'll require the user to specify a return type explicitly.

This applies only to function bodies of function expression-like things, like function expressions, arrow functions, and function shorthand in object literals. I have included many other examples in the linked issue.

marvinhagemeister (approved):
Makes sense, thanks for the further explanation 👍

src/fast_check/mod.rs
```diff
@@ -56,6 +56,36 @@ impl std::fmt::Debug for FastCheckDiagnosticRange {
   }
 }
 
+impl PartialEq for FastCheckDiagnosticRange {
+  fn eq(&self, other: &Self) -> bool {
+    let self_range_start =
+      self.text_info.line_and_column_index(self.range.start);
```
dsherret, line 62: Why does it need to compute the line and column index? Can it just check if the range starts are equal?

src/fast_check/mod.rs
```diff
@@ -56,6 +56,36 @@ impl std::fmt::Debug for FastCheckDiagnosticRange {
...
+      && self_range_end == other_range_end
+  }
+}
+
+impl Eq for FastCheckDiagnosticRange {}
+
+impl std::hash::Hash for FastCheckDiagnosticRange {
+  fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+    let self_range_start =
+      self.text_info.line_and_column_index(self.range.start);
```
dsherret, line 79: Same here.

src/fast_check/mod.rs
```diff
@@ -328,8 +362,14 @@ impl deno_ast::diagnostics::Diagnostic for FastCheckDiagnostic {
...
         Cow::Borrowed("add an explicit type annotation to the symbol")
       }
-      MissingExplicitReturnType { .. } => {
-        Cow::Borrowed("add an explicit return type to the function")
+      MissingExplicitReturnType { is_definitely_void_or_never, is_async, .. } => {
+        if *is_definitely_void_or_never {
+          Cow::Borrowed("add an explicit return type of 'void' or 'never' to the function")
+        } else if *is_async {
+          Cow::Borrowed("add an explicit return type of 'Promise<void>' or 'Promise<never>' to the function")
+        } else {
```
dsherret, line 370: Nice

src/fast_check/mod.rs
```diff
@@ -371,9 +411,18 @@ impl deno_ast::diagnostics::Diagnostic for FastCheckDiagnostic {
...
         Cow::Borrowed("all symbols in the public API must have an explicit type")
       ]),
-      MissingExplicitReturnType { .. } => Cow::Borrowed(&[
-        Cow::Borrowed("all functions in the public API must have an explicit return type")
-      ]),
+      MissingExplicitReturnType { is_definitely_void_or_never, is_async, .. } => {
+        let mut lines = vec![Cow::Borrowed("all functions in the public API must have an explicit return type")];
+        if *is_definitely_void_or_never {
+          if *is_async {
+            lines.push(Cow::Borrowed("async functions without a return statement can have a return type of either 'Promise<void>' or 'Promise<never>'"));
```
dsherret, line 418: Should we clarify this as "function expressions"?

tests/specs/graph/fast_check/expando_props/funcs_ref_private_name.txt
```diff
@@ -8,7 +8,7 @@
 export function test() {
 }
 
-test.skip = (nonExported: NonExported) => {};
```
dsherret, line 11: Why do we add a return statement here?

dsherret (approved):
LGTM, with some questions

src/fast_check/mod.rs
lucacasonato: I guess so - I was somehow thinking that the diagnostic ranges could come from different parses, so the SourcePos would not line up

tests/specs/graph/fast_check/expando_props/funcs_ref_private_name.txt
lucacasonato: Do you want me to change it so we infer an empty function body to `void`?

dsherret: I like how the GH pull request extension caused this to comment on the wrong line (should have been the one below).

It just seems unnecessary (though not a big deal) to have an empty return in the output.

src/fast_check/mod.rs
dsherret: Yeah, I think we can just compare the ranges directly. If they come from different parses they should still be equal.

PR #483, opened 2024-05-22
fix: allow invalid utf8 when decoding modules

sigmaSd:
fixes https://github.com/denoland/deno/issues/23911

bun and node seems to work this way, since they don't error when importing clipper.js (but I haven't  checked their code)

### DIFF
--- a/src/text_encoding.rs
+++ b/src/text_encoding.rs
@@ -31,9 +31,7 @@ pub fn convert_to_utf8<'a>(
   charset: &'_ str,
 ) -> Result<Cow<'a, str>, std::io::Error> {
   match encoding_rs::Encoding::for_label(charset.as_bytes()) {
-    Some(encoding) => encoding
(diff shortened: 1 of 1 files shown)


lucacasonato:
This is correct.

> Let sourceText be the result of [UTF-8 decoding](https://encoding.spec.whatwg.org/#utf-8-decode) bodyBytes.

https://html.spec.whatwg.org/multipage/webappapis.html#fetch-a-single-module-script

> [Process a queue](https://encoding.spec.whatwg.org/#concept-encoding-run) with an instance of [UTF-8](https://encoding.spec.whatwg.org/#utf-8)’s [decoder](https://encoding.spec.whatwg.org/#decoder), ioQueue, output, and "replacement".

https://encoding.spec.whatwg.org/#utf-8-decode

lucacasonato (approved):

PR #512, opened 2024-08-12
 refactor: more specific error handling 

crowlKats:


src/source/file_system.rs
```diff
@@ -11,7 +11,17 @@ pub enum DirEntryKind {
...
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum DirEntryError {
+  #[error("Failed to read directory.")]
+  Dir(std::io::Error),
+  #[error("Failed to read file type.")]
+  FileType(std::io::Error),
+  #[error("Failed to read entry in directory.")]
+  DirEntry(std::io::Error),
```
dsherret, line 24: Do these need to have `#[source]` to say what the error is? Or will thiserror do that automatically?

crowlKats: ah good point, will add that

Cargo.toml
dsherret, line 1: Can you bump file_test_runner to 0.7.4? Also, let me know when this is ready and I'll take a look.

src/graph.rs
```diff
@@ -164,34 +166,50 @@ impl Range {
...
   #[error(transparent)]
-  ContentChecksumIntegrity(ChecksumIntegrityError),
+  ContentChecksumIntegrity(#[inherit] ChecksumIntegrityError),
+  #[class(generic)]
   #[error("Loader should never return an external specifier for a jsr: specifier content load.")]
   ContentLoadExternalSpecifier,
+  #[class(generic)] // TODO: maybe inherit?
   #[error(transparent)]
-  ContentLoad(Arc<anyhow::Error>),
+  ContentLoad(Arc<LoadError>),
```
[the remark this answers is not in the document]
dsherret: Probably it should since it's transparent.

dsherret (approved):
LGTM, but please resolve the todos before merging.

PR #513, opened 2024-08-16
fix: resolve jsxImportSource with types mode in absence of jsxImportSourceTypes

nayeemrmn:
When there is a `jsxImportSource` and no `jsxImportSourceTypes`, we need to try and resolve the same specifier with `ResolutionMode::Types` to see if it resolves somewhere else, as we do with ES deps.

I hit this bug in the LSP specifically when using jsx import source `npm:preact` with byonm.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -2474,6 +2474,27 @@ pub(crate) fn parse_js_module_from_module_info(
             maybe_npm_resolver,
           );
           dep.maybe_deno_types_specifier = Some(specifier_text);
+        } else {
+          let types_resolution = resolve(
+            &specifier_text,
+            range.clone(),
+            ResolutionMode::Types,
+            jsr_url_provider,
+            maybe_resolver,
+            maybe_npm_resolver,
+          );
+          if types_resolution.maybe_specifier()
+            != dep.maybe_code.maybe_specifier()
+          {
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

src/graph.rs
```diff
@@ -2474,6 +2474,27 @@ pub(crate) fn parse_js_module_from_module_info(
...
+            != dep.maybe_code.maybe_specifier()
+          {
+            dep.maybe_type = resolve(
+              &specifier_text,
+              range.clone(),
+              ResolutionMode::Types,
+              jsr_url_provider,
+              maybe_resolver,
+              maybe_npm_resolver,
+            );
```
dsherret, line 2496: Is this just `dep.maybe_type = types_resolution`?

nayeemrmn: Yes, thanks!

dsherret:
It looks like the lockfile changes occurred due to https://github.com/denoland/deno/pull/25059 -- I'm looking into why

dsherret:
Confirmed that https://github.com/denoland/deno_lockfile/pull/34 fixes the issue. I will land this in Deno and then we can merge this PR.

nayeemrmn:
@dsherret Thanks, it's green now

PR #529, opened 2024-09-16
fix: respect @deno-types for type imports

nayeemrmn:
Will fix .. https://github.com/denoland/deno/issues/25644.


src/graph.rs
```diff
@@ -5870,6 +5863,62 @@ mod tests {
...
+    graph.valid().unwrap();
+    let module = graph.get(&Url::parse("file:///foo.ts").unwrap()).unwrap();
+    let module = module.js().unwrap();
+    let dependency = module.dependencies.get("bar").unwrap();
+    assert_eq!(dependency.maybe_code, Resolution::None);
+    assert_eq!(
+      dependency.maybe_type.maybe_specifier().unwrap().as_str(),
+      "file:///bar.d.ts",
+    );
+  }
```
dsherret, line 5920: Note for the future: it's more maintainable to add a test like this in tests/specs/graph probably.

dsherret (approved):
LGTM once main is merged in.

PR #533, opened 2024-09-26
feat: provide full import statement range

nayeemrmn:
Renames `Import::range` to `Import::specifier_range`, adds `Import::full_range`.

Towards https://github.com/denoland/deno/issues/18884.

### DIFF
--- a/src/analyzer.rs
+++ b/src/analyzer.rs
@@ -123,6 +123,8 @@ pub struct StaticDependencyDescriptor {
   pub specifier: String,
   /// The range of the specifier.
   pub specifier_range: PositionRange,
+  /// The range of the import statement.
+  pub range: PositionRange,
   /// Import attributes for this dependency.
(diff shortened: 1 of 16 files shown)


nayeemrmn:
Don't know what caused this:
image

dsherret:
Bug fix: https://github.com/denoland/deno/pull/25860

dsherret (approved):
LGTM

src/analyzer.rs
```diff
@@ -172,6 +174,7 @@ pub struct DynamicDependencyDescriptor {
   pub argument: DynamicArgument,
   /// The range of the argument.
   pub argument_range: PositionRange,
+  pub range: PositionRange,
```
dsherret, line 177: I just realized this breaks the deserializtion of information we store on JSR. We'll need to think of a solution for this.

PR #541, opened 2024-10-25
fix: store resolved root specifiers

nathanwhit:
Basically carry the "rootness" around until the specifier is fully resolved. Once it's resolved, store the resolved specifier in a separate set for later comparison.

This fixes an issue where we would error on `JSON` roots. Ref https://github.com/denoland/deno/issues/26509


dsherret (approved):
Looks great!

PR #554, opened 2024-12-09
feat: add `JsrVersionResolver` trait to allow more control over version resolution

nathanwhit:
this is to support specifying versions to use for jsr packages that are not roots (for `deno outdated --update`). Currently, as a user of deno_graph there isn't a way to influence the version resolution for non-roots, and there also isn't a way to selectively invalidate cached metadata for packages.

### DIFF
--- a/src/packages.rs
+++ b/src/packages.rs
@@ -16,6 +16,16 @@ use serde::Serialize;
 use crate::analyzer::module_graph_1_to_2;
 use crate::ModuleInfo;
 
+pub trait JsrVersionResolver {
+  fn resolve_package_version<'a>(
+    &self,
+    package_info: &'a JsrPackageInfo,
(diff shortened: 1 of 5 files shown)


src/packages.rs
```diff
@@ -16,6 +16,16 @@ use serde::Serialize;
...
 use crate::ModuleInfo;
 
+pub trait JsrVersionResolver {
+  fn resolve_package_version<'a>(
+    &self,
+    package_info: &'a JsrPackageInfo,
+    package_req: &PackageReq,
+  ) -> Option<&'a Version>;
+
+  fn force_fetch_metadata<'a>(&self, package_name: &str) -> bool;
```
nathanwhit, line 26: meh on this, doesn't really feel like it's the right place but adding another trait is annoying

PR #572, opened 2025-02-28
fix: respect lockfile when loading jsr specifiers

nayeemrmn:
Wil fix https://github.com/denoland/vscode_deno/issues/1265#issuecomment-2691214305.


src/graph.rs
```diff
@@ -4158,69 +4148,84 @@ impl<'a, 'graph> Builder<'a, 'graph> {
...
-    }
 
-    None
+      None
+    })()?;
+    let package_nv = PackageNv {
+      name: package_req.name.clone(),
+      version,
+    };
+    self
```
dsherret, line 4224: Here

PR #604, opened 2025-08-20
feat: detect module augmentations as dependencies

nayeemrmn:
Wil fix .. https://github.com/denoland/deno/issues/30444.

Uniquely to this import kind: If the specifier doesn't resolve to some types module being augmented, it won't count as a dependency.


src/ast/dep.rs
```diff
@@ -369,6 +369,32 @@ impl Visit for DependencyCollector<'_> {
...
+            specifier: id_str.value.clone(),
+            specifier_range: id_str.range(),
+            import_attributes: Default::default(),
+            is_side_effect: false,
+          }
+          .into(),
+        );
+      }
+    }
+    node.visit_children_with(self);
```
dsherret, line 396: Do the children need to be visited? Maybe we can skip it? I forget what's allowed in a module augmentation block.

dsherret (approved):
LGTM

src/ast/dep.rs
nayeemrmn: There could be JSDoc `import()`s, at the very least

steveadams:
Thank you @nayeemrmn and @dsherret

Issue #608, opened 2025-09-11
Support re-using `JsrMetadataStore` across `graph.build()` calls


lucacasonato:
Dev server startup on a large Fresh project with a lot of distinct JSR specifiers all resolving to a JSR package with a very large version meta file (>2MB).

Not reusing `JsrMetadataStore`:

<img width="1443" height="747" alt="Image" src="https://github.com/user-attachments/assets/8ba6252b-c32c-47b5-aed3-ba29310a105c" />

Reusing `JsrMetadataStore`:

<img width="1443" height="747" alt="Image" src="https://github.com/user-attachments/assets/d924a64c-5cf2-42c1-8f95-982abb9739f2" />

PR #609, opened 2025-09-11
feat: support re-using JSR metadata store across graph builds

lucacasonato:
Fixes https://github.com/denoland/deno_graph/issues/608

See https://github.com/denoland/deno-js-loader/pull/59


dsherret (approved):
Nice. LGTM!

PR #637, opened 2026-03-12
fix: improve parse error display format

bartlomieju:
## Summary

- Reformats syntax error messages from `ModuleErrorKind::Parse` to use a structured format with location on its own line, pipe margins, line numbers, and carets
- Removes the verbose "The module's source code could not be parsed:" prefix

**Before:**
```
The module's source code could not be parsed: Expected ',', got 'x' at file:///a.js:1:13

  import from "./foo.js"
              ~~~~~~~~~~
```

**After:**
```
Expected ',', got 'x'
    at file:///a.js:1:13
  |
1 | import from "./foo.js"
  |             ~~~~~~~~~~
```

## Test plan
- [x] Updated snapshot test `tests/specs/graph/cjs/file_export.txt`
- [x] Updated JS test assertion in `js/test.ts`
- [x] All 210 deno_graph tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dsherret (commented):
Can you do a ci run with this in the deno repo before merging?

bartlomieju:
Yeah, I opened https://github.com/denoland/deno/pull/32669, but I'm not yet happy with the output

bartlomieju:
PR in Deno is green now: https://github.com/denoland/deno/pull/32669

src/graph.rs
```diff
@@ -562,14 +562,132 @@ impl std::error::Error for ModuleErrorKind {
...
+
+/// Reformats a parse diagnostic message into a nicer format with
+/// line numbers and carets, with the location shown after the snippet.
+///
+/// Input format (from ParseDiagnostic::Display):
+///   `{message} at {specifier}:{line}:{col}\n\n  {source_line}\n  {underline}`
+///
+/// Output format:
+///   `{message}\n  |\nN | {source_line}\n  | {carets}\n    at {location}`
+fn format_parse_diagnostic(
```
dsherret: Is there a better way other than parsing the diagnostic text? This seems very hacky.

dsherret (changes requested):
We should be able to wrap the swc diagnostics with our own errors. I think this is something that should be handled in deno_ast and not here?

PR #649, opened 2026-06-01
feat: type wasm components from wit metadata

divybot:
Adds WebAssembly component and WIT package declaration generation for wasm imports, while preserving core wasm declaration generation for regular modules.\n\nRefs denoland/deno#31314


littledivy:
I'm assuming the test changes are because of the WIT typecheck?

littledivy:
Let's wait for WASI 1.0

PR #651, opened 2026-06-02
feat(source): expose import attribute type in LoadOptions

### LINKED ISSUE #20866: Update `Accept` header for JSON modules with network imports
The HTML spec is being updated to use the appropriate fetch destination for JSON modules: https://github.com/whatwg/html/pull/9486, https://github.com/whatwg/fetch/pull/1691

As far as I can tell, the only relevant different for Deno is that now JSON modules are fetched with the `Accept: application/json,*/*;q=0.5` HTTP header rather than `Accept: */*`. My knowledge of Deno's module system is limited, so please somebody else confirm this.

bartlomieju:
Adds a `maybe_attribute_type` field to the public `LoadOptions` struct that is
passed to `Loader::load` / `Loader::ensure_cached`. It carries the `type` import
attribute the module is being loaded with (e.g. `"json"`, `"text"`, `"bytes"`),
which deno_graph already tracks internally but did not expose to loaders.

This lets a `Loader` set an appropriate `Accept` header on remote requests. In
particular, JSON modules should be fetched with `Accept: application/json,*/*;q=0.5`
per the updated HTML/Fetch specifications (whatwg/html#9486, whatwg/fetch#1691).
The field is populated for the HTTP(S) module load path and left `None` for JSR
registry/content loads where it does not apply. The addition is additive and
does not change the JS loader signature.

Needed by denoland/deno to fix denoland/deno#20866.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -5558,6 +5558,7 @@ impl<'a, 'graph> Builder<'a, 'graph> {
           was_dynamic_root: self.was_dynamic_root,
           cache_setting: CacheSetting::Only,
           maybe_checksum: Some(checksum.clone()),
+          maybe_attribute_type: None,
         },
       );
       let is_dynamic_branch = self.in_dynamic_branch;
@@ -5965,6 +5966,9 @@ impl<'a, 'graph> Builder<'a, 'graph> {
           was_dynamic_root,
           cache_setting: CacheSetting::Use,
           maybe_checksum: maybe_checksum.clone(),
+          maybe_attribute_type: maybe_attribute_type
+            .as_ref()
+            .map(|t| t.kind.clone()),
         };
 
         let handle_redirect =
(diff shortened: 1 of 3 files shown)


dsherret:
We agreed to not bother doing this earlier and this is not the right fix. If you implement this properly, then you need to start keying modules on specifier and import attribute, which introduces significant complexity. What happens if someone loads a module with a json import attribute and then another module without one and then what happens if the server responds differently based on the accept header? Ex. what if the server loads a js file at one and a json file at the other? Then, what do you do if someone loads that specifier in the LSP? Which module do you load? Modules are no longer keyed on just specifier, which is just too complicated—don't bother.

bartlomieju:
Agree, that's too niche of a use case. That said, I might still try to tackle that for the module loader hooks problems I was running into (see deferred for follow ups: https://github.com/denoland/deno/pull/34390#issuecomment-4605936660)

PR #658, opened 2026-07-09
fix: prefer cached jsr versions when resolving under cached-only

magurotuna:
First half of the fix for denoland/deno#35879 (`deno run --cached-only` can fail after a successful `deno cache` due to JSR resolution divergence with type-only imports). The second half is a small deno CLI change that sets the new flag when running with `--cached-only`.

## Problem

`deno cache` resolves the type-inclusive graph; `deno run` (no type-checking) resolves a code-only graph. When an exact JSR pin is reachable **only** through a type-only import and another dependency requires a semver range on the same package, the cache-building command unifies the range onto the pinned (older) version and materializes only that version. The runtime graph doesn't contain the pin, re-resolves the range against the cached package `meta.json` to the newest matching version, and fails under `--cached-only` because that version's manifest was never fetched:

```
JSR package version manifest for '@std/encoding@1.0.11' failed to load:
Specifier not found in cache: "https://jsr.io/@std/encoding/1.0.11_meta.json", --cached-only is specified.
```

(Real-world instance in denoland/deno#35879: Lume pins `jsr:@std/encoding@1.0.10` on a type-only path while `jsr:@std/http@1.0.22/etag.ts` requires `^1.0.10`; the moment `1.0.11` was published, every lockfile-less build/run pair broke with no change to the project.)

## Change

Adds an opt-in `BuildOptions::prefer_cached_jsr_versions` (default `false`, so nothing changes for existing consumers). When enabled, resolving a `jsr:` semver range inserts a new step **1.5** into `JsrPackageVersionResolver::resolve_version`, between in-graph unification (step 1) and unyanked registry versions (step 2): resolve the range against the set of matching, unyanked versions whose `<version>_meta.json` is already in the cache (probed via `Loader::load` with `CacheSetting::Only`, which by contract returns `Ok(None)` when not cached).

The placement makes the flag strictly error-converting for cached-only consumers (rationale also in a code comment):

- Step 1 stays first, so in-graph unification still wins — matching online behavior when a code-edge pin exists in the graph.
- 1.5 is not merged into step 1's candidate iterator: if the graph unified onto an older version while a newer matching version is also cached, chaining would diverge from what an online run resolves.
- Before step 2, the change can only convert errors into successes: if step 2's best pick is cached it equals the best cached match here; if it isn't cached, step 2's pick fails to load today.
- Yanked versions are never selected from the cached set (consistent with step 2 preceding step 3).

## Tests

- `tests/specs/graph/jsr/prefer_cached_version.txt` — `meta.json` lists 1.0.10/1.0.11, only 1.0.10's manifest cached, flag on → range resolves to 1.0.10.
- `tests/specs/graph/jsr/prefer_cached_version_off.txt` — same fixture, flag off → today's behavior (resolves 1.0.11 and fails to load it).
- `tests/specs/graph/jsr/prefer_cached_version_unification.txt` — code-edge exact pin + range with **both** versions cached, flag on → unification wins (both resolve to the pin).

`cargo test --locked --all-features --all-targets`, `cargo fmt --check`, and `cargo clippy --locked --all-features --all-targets -- -D clippy::all` are green.

## Notes for review

- The cached-manifest probe runs per pending req when the flag is on (one cache-only load per matching version), even when step 1 unification would decide. Probes are local cache reads, so this is kept simple; it could be skipped when the graph already has a matching selection or memoized per package if you'd prefer.
- `newest_dependency_date` is applied in step 1.5 the same way as step 2 (unlike step 1, which ignores it), keeping 1.5's candidate set a strict subset of step 2's.

### DIFF
--- a/lib/lib.rs
+++ b/lib/lib.rs
@@ -315,6 +315,7 @@ pub async fn js_create_graph(
         unstable_text_imports: true,
         unstable_css_imports: true,
         unstable_config_imports: true,
+        prefer_cached_jsr_versions: false,
         resolver: maybe_resolver.as_ref().map(|r| r as &dyn Resolver),
         // todo(dsherret): actually implement this for Wasm users
         // and don't just use a RealSys here as it would be better
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -53,6 +53,7 @@ use deno_semver::package::PackageReq;
 use deno_semver::package::PackageReqReferenceParseError;
 use futures::FutureExt;
 use futures::future::LocalBoxFuture;
+use futures::future::join_all;
 use futures::stream::FuturesOrdered;
 use futures::stream::FuturesUnordered;
 use futures::stream::StreamExt;
@@ -4532,6 +4533,20 @@ struct PendingJsrReqResolutionItem {
   is_root: bool,
 }
 
+/// Memoizes the results of cache-only JSR version-manifest probes for a
+/// single package within one `resolve_pending_jsr_specifiers` pass, so that
+/// multiple pending reqs on the same package don't repeat the (cache-only)
+/// manifest existence checks. Only populated when `prefer_cached_jsr_versions`
+/// is enabled.
+#[derive(Debug, Default)]
+struct CachedJsrVersionProbe {
+  /// Versions whose manifest cache presence has already been checked in this
+  /// pass (regardless of whether it was found in the cache).
+  probed: HashSet<Version>,
+  /// Subset of `probed` whose version manifest was found in the cache.
+  cached: HashSet<Version>,
+}
+
 #[derive(Debug)]
 struct PendingJsrNvResolutionItem {
   specifier: ModuleSpecifier,
@@ -4899,6 +4914,15 @@ impl<'a, 'graph> Builder<'a, 'graph> {
         && self.graph.graph_kind.include_types();
     // don't bother pre-allocating because adding to this should be rare
     let mut restarted_pkgs = HashSet::new();
+    // Memoizes cache-only version-manifest probe results per package name for
+    // the duration of this pass so multiple pending reqs on the same package
+    // don't repeat probes. Only populated when `prefer_cached_jsr_versions` is
+    // enabled and in-graph unification isn't going to decide the version.
+    let empty_cached_versions = HashSet::new();
+    let mut cached_versions_by_package: HashMap<
+      StackString,
+      CachedJsrVersionProbe,
+    > = HashMap::new();
     while let Some(pending_resolution) = pending_resolutions.pop_front() {
       let package_name = &pending_resolution.package_ref.req().name;
       let fut = self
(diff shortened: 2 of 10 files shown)


magurotuna:
CI notes:

- The `javascript` job failure was real: the wasm binding's exhaustive `BuildOptions` initializer in `lib/lib.rs` was missing the new field (it's outside the default workspace build, so `cargo test --all-targets` didn't catch it). Fixed in a982b2e.
- The `shard 0` / `shard 4` ecosystem failures (`ecosystem::bureaudouble::sonner::0_0_1`, `ecosystem::ipinfo::client::1_0_1`) look like canary drift rather than this PR: they fail in the TYPE CHECK phase (`deno check --no-config` under `deno-version: canary`) with DOM-lib type errors (`Cannot find name 'HTMLButtonElement'`), and this PR's only behavioral surface is JSR version selection behind a default-off flag — it doesn't touch type checking. main last ran green on 2026-07-06 with that day's canary; this run used 2026-07-09's. Happy to be corrected if a maintainer knows otherwise — these may need the usual "update ecosystem test expectations for current canary" chore.

magurotuna:
Confirmed the ecosystem failures are pre-existing canary drift, not this PR: running the **unmodified `main` (ab72f6e)** through the ecosystem workflow on a fork against today's canary fails the exact same three specs — `ecosystem::bureaudouble::sonner::0_0_1` (shard 0), `ecosystem::ipinfo::client::1_0_1` (shard 4), `ecosystem::mebus::mebus::1_0_1` (shard 5): https://github.com/magurotuna/deno_graph/actions/runs/29004670090 (the shard 7 failure there is fork-only infrastructure — cold JSR-mirror cache download flake).

These need the usual "update ecosystem test expectations for current canary" chore (like #654, which also touched `ipinfo/client`); happy to include the refreshed snapshots here or keep them in a separate chore PR, whichever is preferred.

bartlomieju:
On the cached-manifest probe (the "Notes for review" bullet): I'd lean toward tightening this before merge, since as written it's a latent perf cliff rather than just a micro-cost.

`load_cached_jsr_version_manifests` issues one `CacheSetting::Only` load **per matching version**, awaited **sequentially**, and runs for **every** pending JSR req whenever the flag is on — including when step-1 in-graph unification is going to decide the result and throw the probe away.

The cost scales with published version count, not with anything the resolution actually needs:

- A range like `^1.0.10` against a package with a long patch history probes every matching `<version>_meta.json`, one `load().await` at a time.
- When an in-graph pin already exists, `resolve_jsr_nv` returns from step 1 before `cached_versions` is ever read, so the entire probe is pure waste.
- Each probed `_meta.json` for the version that ultimately wins is then loaded a second time by `queue_load_package_version_info`.

Since the flag is meant to be on by default under `--cached-only`, this is on the hot path for the exact command it targets. Three cheap mitigations, roughly in order of impact:

1. **Skip when unification will win** — if `self.graph.packages.versions_by_name(&package_req.name)` already yields a version matching `version_req`, step 1 decides and the probe is unnecessary; bail before probing.
2. **Parallelize** — `join_all` the per-version loads instead of the serial `for` loop; they're independent cache reads.
3. **Memoize per package name** across pending reqs in a single `resolve_pending_jsr_specifiers` pass, so N reqs on the same package don't re-probe.

(1) alone removes the cliff for the common unified case; (1)+(2) keep the worst case bounded. Happy to be wrong if the version counts in practice are always small, but the unbounded-with-history shape is what makes me want to guard it rather than leave it simple.

magurotuna:
@bartlomieju thanks, this makes sense — tightened in 927c3404.

What changed:
- Skip the cached-manifest probe when the resolver's existing in-graph unification pass will decide the version. The skip mirrors that pass: it only checks already-selected graph versions with `version_req.matches(...)`, without applying `newest_dependency_date`, so resolution output should remain unchanged.
- Memoize cached version-manifest probe results per package name within a single pending JSR resolution pass, so overlapping reqs for the same package don’t re-probe the same manifests.
- Parallelize the independent cache-only manifest loads with `join_all`.
- Added test-only probe counting plus count-based regression tests for the unification-skip and per-package memoization cases.

For the probe-count checks I used synthetic cases matching the specific cliffs from the review:

- Unification-wins case: `@scope/a` has 10 cached matching manifests, but the graph/lockfile already has an in-graph selection satisfying the req. Before, all 10 manifests were probed before the in-graph unification pass discarded the result; after, this is `10 -> 0`.
- Same-package duplicate reqs: one req matched 10 versions (`^1.0.0`) and another overlapping req matched 5 of those (`~1.0.0`). Before, the overlapping versions were re-probed (`10 + 5 = 15`); after, the per-package memo avoids re-probing and total probes are `10`.
- Range-only case: 100 matching versions with no unification. Probe count is necessarily still `100`, but a throwaway smoke check showed the per-version cache-only loads are no longer sequential: max concurrent probes `1 -> 100`.

Verified with fmt, diff-check, focused tests, full `--all-features --all-targets`, and clippy:

- `cargo fmt --check`
- `git diff --check`
- `cargo test --locked prefer_cached_jsr_versions`
- `cargo test --locked --test integration_test`
- `cargo test --locked --test specs`
- `cargo test --locked --all-features --all-targets`
- `cargo clippy --locked --all-targets --all-features -- -D clippy::all`

I intentionally didn’t add a persistent concurrency/timing test for the parallelization part because the test loader is in-memory and delay-based tests would be brittle; the durable regression tests are count-based.

PR #661, opened 2026-08-21
feat: resolve wildcard requirements to the newest prerelease for packages with only prerelease versions

crowlKats:
A `*` version requirement never matches a prerelease version, so a JSR package that has no stable release could never be resolved by `deno run jsr:@scope/pkg`, `deno add jsr:@scope/pkg`, etc. — the requirement matched nothing and resolution failed.

This mirrors npm's behavior for such packages, where `*` resolves to the `latest` dist-tag even when that is a prerelease.

## Behavior

Resolution additionally considers prerelease versions when **both** hold:

1. the requirement is a wildcard — `*`, `x`, `X`, or absent, as in `jsr:@scope/pkg`; and
2. the package has no *usable* stable release.

"Usable" excludes yanked versions: a package whose only stable release was yanked is in the same position as one that never published a stable release, and npm's `latest` points at the prerelease in that case too.

Everything else is unchanged. A constrained requirement (`^1.0.0`) never takes the fallback and still fails against a prerelease-only package. Versions excluded by the newest dependency date stay excluded, and the error still reports that date. Yanked prereleases are still only reached after unyanked ones, and still mark the graph as using a yanked package.

## Implementation

The decision is made once, up front, from the package's own versions:

```rust
pub fn prerelease_fallback_applies(
  version_req: &VersionReq,
  package_info: &JsrPackageInfo,
) -> bool {
  *version_req == *WILDCARD_VERSION_REQ
    && !package_info
      .versions
      .iter()
      .any(|(version, info)| version.pre.is_empty() && !info.yanked)
}
```

Comparing against `WILDCARD_VERSION_REQ` rather than the requirement's raw text means `x` and `X` are handled, since `deno_semver` parses all three spellings to the same range and compares `VersionReq` by parsed value.

The resulting flag is applied uniformly across all four steps of `resolve_version` (existing versions, cached versions, unyanked versions, yanked versions), so the fallback is a property of the package rather than of a particular step or a failed attempt.

`version_matches` is the single definition of "is this version a candidate", and `graph.rs` shares it. Without that, `prefer_cached_jsr_versions` would probe only the versions the requirement literally matches — none, for a prerelease-only package — and then resolve to a version whose manifest is not in the cache, defeating the point of the flag.

## Breaking change

`ResolveVersionOptions` gains a required public field, `include_all_prereleases`, so downstream struct literals stop compiling. **Release this as `minor`, not `patch`.**

The field deliberately bypasses the version requirement, which its docs say outright; derive it with `prerelease_fallback_applies` rather than setting it by hand.

## Tests

- resolves to the newest prerelease for a prerelease-only package
- the same for the `x` and `X` wildcard spellings
- a constrained requirement still fails, asserted on `JsrLoadError::PackageReqNotFound` rather than bare `is_err()`
- a package with a usable stable release resolves to it, not to a higher prerelease
- a yanked stable release does not suppress the fallback, and the graph is not marked as using a yanked package
- the newest dependency date still excludes prereleases, falling back to the newest one published before the cutoff
- when every prerelease is excluded by the date, the error reports the date
- `prefer_cached_jsr_versions` probes the prereleases the fallback will consider, and prefers the cached one

## Not addressed here

A lockfile pin of a wildcard requirement to a prerelease is dropped once upstream publishes a stable release (#666). It shares a code path with this change but is pre-existing, and the one-line fix — letting step 1 match prereleases for wildcard requirements — would also make a bare `jsr:@scope/pkg` unify onto a prerelease whenever any other import in the graph opted into one. That needs step 1 to distinguish a lockfile pin from a version resolved a moment earlier in the same pass.

---

Counterpart to the registry-side change in jsr-io/jsr#1497; the CLI side is in denoland/deno. Ref jsr-io/jsr#897.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -27,6 +27,8 @@ use crate::module_specifier::resolve_import;
 use crate::packages::JsrPackageInfo;
 use crate::packages::JsrPackageVersionInfo;
 use crate::packages::PackageSpecifiers;
+use crate::packages::prerelease_fallback_applies;
+use crate::packages::version_matches;
 use crate::rt::Executor;
 
 use crate::source::*;
@@ -4935,7 +4937,7 @@ impl<'a, 'graph> Builder<'a, 'graph> {
         Ok(info) => {
           let package_req = pending_resolution.package_ref.req();
           let cached_versions = if !self.prefer_cached_jsr_versions
-            || self.jsr_unification_decides(package_req)
+            || self.jsr_unification_decides(package_req, &info)
           {
             // Either the feature is off, or in-graph version unification
             // (step 1 of `resolve_version`) is going to decide this req, in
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):
LGTM

PR #662, opened 2026-08-22
fix(fast_check): infer literal types for substitution-less template literals

### LINKED ISSUE #453: Untagged template literals are incorrectly marked as slow types
```
// these should be fine
const x = `hello world`;
const x = `hello ${foobar} world`;
const x = `hello world` as const;

// these are not allowed
const x = tpl`hello world`;
const x = `hello ${foobar} world` as const;
```

### LINKED ISSUE #458: Allow expr-less template literals in const positions in slow types
```ts
const x = `asd`;
let x = `asd` as const;`
export default `asd`;
```

not allowed:

```ts
const x = `asd ${asd}`;
```

crowlKats:
Untagged template literals were always inferred as `string`, but in a const position TypeScript infers the string literal type:

| source | tsc | before |
| --- | --- | --- |
| ``const a = `hello world` `` | `"hello world"` | `string` |
| ``const b = `hello world` as const`` | `"hello world"` | `string` (dts) |
| ``let c = `hello world` `` | `string` | `string` |
| ``const d = `hi ${foobar}` `` | `string` | `string` |
| ``readonly e = `hello` `` (class prop) | `"hello"` | `string` |

Adds `tpl_to_ts_type()` in `swc_helpers.rs`, used by both the fast check transform and the dts transform, which returns the string literal type of the cooked value for a substitution-free template in a const position and `string` otherwise.

Template literals with substitutions keep producing `string` (matching TypeScript), and tagged template literals remain slow types.

One known gap: the `.d.ts` for a const-asserted template *with* substitutions still widens to `string`, since the real type (`` `hi ${string}` ``) can't be computed without a type checker. The fast check `.ts` emit leaves the expression as-is, so it stays correct there.

Closes https://github.com/jsr-io/jsr/issues/453
Closes #458

### DIFF
--- a/src/fast_check/swc_helpers.rs
+++ b/src/fast_check/swc_helpers.rs
@@ -184,6 +184,29 @@ pub enum DeclMutabilityKind {
   Mutable,
 }
 
+/// Gets the type of an untagged template literal expression.
+///
+/// A template literal without substitutions is equivalent to a string
+/// literal, so it keeps its literal type in a const position. Otherwise
+/// the type depends on the substituted expressions, which TypeScript only
+/// evaluates in a const assertion (that case is handled by leaving the
+/// expression as-is instead of inferring a type from it).
+pub fn tpl_to_ts_type(tpl: &Tpl, decl_kind: DeclMutabilityKind) -> TsType {
+  if let DeclMutabilityKind::Const = decl_kind
+    && tpl.exprs.is_empty()
+    && let Some(quasi) = tpl.quasis.first()
+    && let Some(cooked) = &quasi.cooked
+  {
+    return ts_lit_type(TsLit::Str(Str {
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -76,4 +76,4 @@ jobs:
       - name: Test
         run: deno task test
       - name: Publish JSR
-        run: cd js && deno run -A jsr:@david/publish-on-tag@0.1.3
+        run: cd js && deno run -A --no-lock jsr:@david/publish-on-tag@0.1.3
```
dsherret, line 79: Maybe move this dep into deno.json and then do `deno run -A @david/publish-on-tag` here. That way it's locked in the lockfile.

bartlomieju (approved):

PR #663, opened 2026-08-22
chore: update ecosystem test expectations for native tsc

crowlKats:
The ecosystem shards have been red on main — and therefore on every PR — since mid-July. Cause: `deno check` switched from the forked in-isolate TypeScript to the native compiler (`typescript@7` / tsgo) in denoland/deno#35938, merged 2026-07-17. The last green ecosystem run here was 2026-07-09, eight days earlier, so every expectation in `tests/specs/ecosystem/` was recorded with the old checker.

Confirmed it is not any one PR: an empty commit on main (`ci_control_empty`, zero code changes) fails with the same 923 specs as #661 and #662, shard for shard.

Regenerated against **deno 2.9.5+98f9507 (canary)** with `UPDATE=1 cargo test --test ecosystem`, then verified with a plain `cargo test --test ecosystem` — **7555 passed**.

Diagnostics moved in both directions:

| | packages |
| --- | --- |
| failed → passed | 9 |
| passed → failed | 592 |

Dominant new codes among the regressions: TS2694 (201), TS5089 (90), TS2846 (38), TS2709 (33), TS2305 (29), TS2306 (24).

Two things worth a look before this is treated as settled:

- **TS2846** — "A declaration file cannot be imported without 'import type'" — fires on ordinary Deno code that the old checker accepted. Minimal repro: `import { Foo } from "./types.d.ts"` type-checks clean on deno 2.9.1 stable and fails on canary. If that is an unintended tsgo behaviour, these 38 expectations encode a bug.
- The native compiler version is pinned by deno, so this will need regenerating again as canary moves. Pinning `deno-version` in `ecosystem.yml` to a fixed build would make the suite reproducible, at the cost of no longer catching canary breakage early.


crowlKats:
Parking this. #664 pins the ecosystem workflow to an exact stable Deno release instead of `canary`, which unblocks CI now — the native-tsc switch (denoland/deno#35938) has not reached a stable release yet, so the current expectations on `main` still pass.

This branch keeps the regenerated-for-native-tsc expectations for whenever that behaviour does ship in stable. It will need a fresh `UPDATE=1` regen at that point rather than a merge as-is, since the exact diagnostics will have moved again.

PR #667, opened 2026-08-26
fix(fast_check): improve type inference for dts generation and support Object.freeze / new expressions

crowlKats:
A sweep of the jsr slow-types / d.ts generation issue backlog surfaced several fast check inference gaps. Each commit is self-contained:

**fix: infer types for identifier references in dts** (jsr-io/jsr#615)
An object property referencing a function (`export const f = { add: fAdd }`) emitted `readonly add;` with no type annotation in the generated .d.ts, so npm consumers got `any`. Identifier references now emit `typeof ident` when the referenced binding has a non-widening declared type (function/class declarations, consts with an explicit annotation or function initializer — tracked via a prescan of the module). Shorthand properties are supported the same way, global `undefined`/`NaN`/`Infinity` are inferred, and uninferable properties now fall back to an explicit `any` with a diagnostic instead of an implicit-any property signature.

**fix: keep well-known symbol members in dts output** (jsr-io/jsr#886, jsr-io/jsr#1179)
Class members keyed by well-known symbols (`[Symbol.iterator]`, `[Symbol.dispose]`, `[Symbol.toStringTag]`, …) were silently dropped from the generated .d.ts, breaking `Iterable`/`Disposable` implementations for npm consumers. They are now kept when `Symbol` is the global (a shadowed local `Symbol` still drops them).

**fix: support spreads and negative number literals in dts tuple inference** (jsr-io/jsr#1258)
`[...OTHER, "a"] as const` silently dropped the spread elements from the emitted tuple type (grammY's `ALL_UPDATE_TYPES` shipped as a 3-element tuple instead of 22). Spreads of identifiers now emit `...typeof ident`, spreads of array literals emit a rest of their tuple type, and other spreads give up on the whole array (any + diagnostic) instead of emitting a tuple with missing elements. Unary-minus number/bigint literals are now inferred (`-2` was silently dropped from tuples), and uninferable elements keep their position as `any`.

**feat: support Object.freeze in the public API** (jsr-io/jsr#155)
`export const prefixes = Object.freeze({ ... } as const satisfies Record<string, number>)` no longer errors with missing-explicit-type: `Object.freeze(x)` is leavable when its argument is, and the dts transform emits `Readonly<T>` of the inferred argument type.

**feat: support new expressions of non-generic module-level classes** (jsr-io/jsr#155)
`new Dimensions(...)` (a class declared in the same module without type parameters) is now leavable and infers the class instance type in the dts transform, which the second real-world constant in jsr-io/jsr#155 needed. Generic classes still require an explicit annotation since their type arguments can't be inferred without a checker.

The ecosystem expectations update removes ~700 lines of now-resolved diagnostics from real packages (noble/ed25519's `static readonly ZERO = new Point(...)`, denosaurs/python's `new Python()`, bradenmacdonald/quantity-math-js's frozen unit tables, and vxern/dexonline-scraper now fast-checks entirely instead of failing).


src/fast_check/transform_dts.rs
```diff
@@ -311,16 +341,64 @@ impl<'a> FastCheckDtsTransformer<'a> {
...
-              as_readonly,
-            ) {
-              Some(ts_expr) => {
-                elem_types.push(ts_tuple_element(ts_expr));
+            if expr_or_spread.spread.is_some() {
+              // tuple types support spreading array-likes, so a spread of
+              // an identifier can be represented as `...typeof ident` and
+              // a spread of an array literal as a rest of its tuple type
+              let rest_type = match &*expr_or_spread.expr {
+                Expr::Ident(ident) if ident.ctxt != self.unresolved_context => {
```
bartlomieju: Could we guard this to identifiers whose inferred type is known to be an array or tuple? For example, `const source: string = "abc"; export const result = [...source] as const;` is valid source, but this emits `readonly [...typeof source]`, which TypeScript rejects with TS2574 because a tuple rest type must be an array type. Other non-array iterables such as `Set<T>` have the same problem. If the shape cannot be established syntactically, falling back with a diagnostic would avoid producing an invalid declaration.

src/fast_check/transform_dts.rs
```diff
@@ -477,20 +587,66 @@ impl<'a> FastCheckDtsTransformer<'a> {
...
+        _ => None,
+      },
+      Expr::Call(call_expr) => {
+        // `Object.freeze(expr)` returns `Readonly<T>` of its argument
+        let arg = object_freeze_call_arg(&call_expr, self.unresolved_context)?;
+        let inner_type =
+          self.expr_to_ts_type((*arg.expr).clone(), as_const, as_readonly)?;
+        Some(TsType::TsTypeRef(TsTypeRef {
+          span: DUMMY_SP,
+          type_name: TsEntityName::Ident(new_ident("Readonly".into())),
```
bartlomieju: This unqualified `Readonly` can bind to a module-local type declaration. For example, with `export type Readonly<T> = string; export const frozen = Object.freeze({ value: 1 });`, the emitted type resolves to the exported alias and makes `frozen` a `string`. Could this emit an equivalent mapped/structural type that cannot be shadowed?

bartlomieju (approved):
Looks good overall. I left two inline comments for declaration-generation edge cases that would be useful to cover.

src/fast_check/transform_dts.rs
crowlKats: Good catch — fixed in 66457316. The prescan now tracks bindings whose type is provably an array/tuple (array or tuple annotation, optionally `readonly`-wrapped, `Array`/`ReadonlyArray` refs, or an array-literal initializer for consts), and `...typeof ident` is only emitted for those. Anything else — including the `[...source]`-of-a-string case — falls back to `any` with the spread diagnostic. Added a unit test with exactly that repro.

src/fast_check/transform_dts.rs
crowlKats: Fixed in 66457316. `Object.freeze` now emits a structural equivalent instead: type-literal members get `readonly` applied directly, arrays/tuples get the `readonly` operator, function types pass through unchanged (`freeze<T extends Function>(f: T): T`), and shapes that can't be expressed structurally fall back to `any` with a diagnostic rather than emitting a mapped type (which would drop call signatures). Added a unit test with a module-local `export type Readonly<T> = string;` shadow.

PR #668, opened 2026-09-05
fix: resolve triple-slash path references relatively

cqxswbc2:
## Summary

- resolve triple-slash `path` references relative to the containing module even when they omit `./`
- keep the original source specifier in dependency metadata and diagnostics while normalizing only the value passed to resolvers
- add a regression test using a custom resolver, matching the path that exposed denoland/deno#36751

This is the upstream graph fix for denoland/deno#36751. A Deno dependency bump will still be needed after this is released.

## Tests

- `cargo fmt --check`
- `cargo clippy --locked --all-features --all-targets -- -D clippy::all`
- `cargo test --locked --all-features --all-targets` (156 unit, 24 integration, and 228 spec tests passed; ecosystem mirror test skipped because no local mirror was configured)
- `cargo build --locked --no-default-features`

## AI assistance disclosure

Codex was used to investigate the bug, implement the change, run the test suite, and perform an independent adversarial review. The submitted diff passed the repository's Rust CI-equivalent checks locally.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -2944,10 +2944,56 @@ fn resolve(
   jsr_url_provider: &dyn JsrUrlProvider,
   maybe_resolver: Option<&dyn Resolver>,
 ) -> Resolution {
-  let response = if let Some(resolver) = maybe_resolver {
-    resolver.resolve(specifier_text, &referrer_range, resolution_kind)
+  resolve_with_specifier_text(
+    specifier_text,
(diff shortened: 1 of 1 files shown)


cqxswbc2:
Following up on this PR: the cross-platform and ecosystem checks are green and the CLA is complete. Please let me know if any implementation detail or test case needs adjustment.

